PR #4, opened 2020-06-05
Prep for release

72636c:
This should be version-matched with the old scoped package.


package.json
```diff
@@ -7,28 +7,30 @@
...
   "module": "lib/es2015",
   "typings": "lib/index.d.ts",
   "files": [
-    "config",
-    "lib",
+    "config/**/*",
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "lib/**/*.js.map",
+    "lib/template/**/*",
```
72636c, line 19: We were previously bundling `tsbuildinfo`s 🤦

This chicanery appears to get us down from ~1.5MB to ~0.5MB

PR #9, opened 2020-06-05
Test templates in CI

72636c:


72636c:
Requires https://github.com/seek-oss/skuba-dive/pull/4

.github/workflows/validate.yml
```diff
@@ -23,10 +22,67 @@ jobs:
...
+      - name: Set up Node.js 12.x
+        uses: actions/setup-node@master
+        with:
+          node-version: 12.x
+
+      - name: Test template
+        run: yarn test:template greeter
+
+  koa-rest-api:
+    name: koa-rest-api
```
tobyhei, line 53: can each template use an actions matrix? (I've never used GH actions so don't really know what I'm talking about) https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

72636c: that sounds fancy so I'll check it out

72636c: works a treat 👏

PR #11, opened 2020-06-05
Flatten skuba bundle

72636c:
A prominent pain point I've observed working with seek-module-toolkit is that it compiles to nested directories like `/lib/commonjs`. This means that the depth of the compiled file differs from the source file under `/src`, which can make bundling and referencing of non-JS assets more of a hassle than it should be.

I propose that we instead compile to `/lib-commonjs`, and document this as a potential breaking change when migrating from seek-module-toolkit. skuba itself moves to this model as a guinea pig.

I've taken this opportunity to revise the ignore files to account for proposed directories like `/lib-commonjs`.


72636c:
Note that this doesn't actually implement the migration path from seek-module-toolkit; that's tracked in #13.

PR #20, opened 2020-06-06
Exclude files from templating based on .gitignore

72636c:
We used to exclude a hardcoded list of "filenames" from templating, so we didn't chug on trying to rewrite all of `.git` or `node_modules`. This easily fell out of sync and couldn't represent complex paths.

Instead, we use `.gitignore`s as a basis for which files to exclude.


src/cli/init/prompts.ts
```diff
@@ -46,24 +46,3 @@ export const TEMPLATE_PROMPT = new Select({
   message: 'Select a template:',
   name: 'templateName',
 });
-
-export const IGNORED_TEMPLATE_FILENAMES = new Set([
```
etaoins, line 50: 🥇

PR #34, opened 2020-06-10
Add Assert.notNullish function

72636c:
This seems trivial enough that it wouldn't hurt to be bundled in skuba.

### DIFF
--- a/.changeset/hungry-ghosts-compare.md
+++ b/.changeset/hungry-ghosts-compare.md
@@ -1,0 +1,7 @@
+---
+'skuba': minor
+---
+
+Add `Assert.notNullish` function
+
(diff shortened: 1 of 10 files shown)


72636c:
Ooh good point. We've only ever used these in tests so I added it here, but it certainly could be used in actual production code 🤔

I've been a bit iffy on how much validation to put into `skuba-dive`, but maybe simple assertions like this make sense?

tobyhei:
I'm not overly fussed, it was more me trying to get a better understanding of the purpose of `skuba dive`

72636c:
I think it probably makes more sense there. `skuba-dive` is meant to be a dumping ground that's not a dumping ground 😅

PR #44, opened 2020-06-12
Update Koa template for latest Koala

72636c:


template/koa-rest-api/src/testing/server.ts
```diff
@@ -15,12 +15,12 @@ export const agentFromApp = <State, Context>(app: Koa<State, Context>) => {
...
 
-  const setup = (done: () => void) => {
-    server = app.listen(undefined, done);
+  const setup = async () => {
+    await new Promise((resolve) => (server = app.listen(undefined, resolve)));
     agent = request.agent(server);
   };
 
-  const teardown = (done: () => void) => server.close(done);
+  const teardown = () => new Promise((resolve) => server.close(resolve));
```
72636c, line 23: This is mildly unrelated but I noticed it when working on Koala; the promises are a bit more modern, and it means that you can `await teardown()` along with other things rather than hogging Jest's `done` callback.

PR #51, opened 2020-06-14
Git more on init

72636c:
We can automatically configure the remote since we now capture GitHub org and repo details. Making some initial commits is another easy win.

### DIFF
--- a/src/cli/init/git.ts
+++ b/src/cli/init/git.ts
@@ -2,8 +2,61 @@ import path from 'path';
 
 import fs from 'fs-extra';
 
-import { createExec } from '../../utils/exec';
+import { Exec, createExec } from '../../utils/exec';
 import { log } from '../../utils/logging';
+
+interface GitHubProject {
+  orgName: string;
+  repoName: string;
+}
+
+export const initialiseRepo = async (
+  exec: Exec,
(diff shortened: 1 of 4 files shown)


src/cli/init/git.ts
```diff
@@ -2,9 +2,62 @@ import path from 'path';
...
+  await exec(
+    'git',
+    '-c',
+    'user.email=<>',
+    '-c',
+    'user.name=skuba',
+    'commit',
+    '--allow-empty',
+    '--author',
+    'skuba <>',
```
72636c, line 28: This is hyper-defensive so we don't blow up if you haven't configured your Git username or email 🤷

src/cli/init/git.ts
```diff
@@ -2,9 +2,62 @@ import path from 'path';
...
+    'git',
+    'remote',
+    'add',
+    'origin',
+    `git@github.com:${orgName}/${repoName}.git`,
+  );
+};
+
+export const commitChanges = async (exec: Exec, message: string) => {
+  await exec('git', 'add', '--all');
```
etaoins, line 44: It's annoying that `git commit -a` nearly-but-not-quite works here

PR #53, opened 2020-06-14
Refactor more configure code

72636c:
This sets us up to configure dependencies.


src/cli/configure/analyseConfiguration.ts
```diff
@@ -0,0 +1,48 @@
...
+import { log } from '../../utils/logging';
+
+import { diffFiles } from './analysis/project';
+
+interface Props {
+  destinationRoot: string;
+  entryPoint: string;
+}
+
+export const analyseConfiguration = async (
```
72636c, line 14: This is just a cut down version of `src/cli/configure/applyConfiguration.ts` as some logic was lifted to the `index.ts` function

PR #54, opened 2020-06-14
Configure dependencies

72636c:
This aims to ease the migration from our old internal packages to their OSS successors.

Builds on #53.

---

Trial run on an internal repo (_ignore_ the unrelated ignore file churn due to #11):

Screen Shot 2020-06-14 at 5 03 41 pm


72636c:
Good question. I'm thinking that we release the next version across both packages, then formally deprecate `@seek/skuba`. A major benefit is that ESLint 7 changes can be resolved as part of the regular Renovate flow rather than making it a barrier to migration.

In practice, the laziest option will be to `yarn global add skuba` once and then `skuba configure` each repo. I'll have some doco once we're closer to release.

PR #58, opened 2020-06-16
Update skuba managed sections when merging ignore files

lumishrestha:
Added skuba managed section to ignore files in the base template. So,`skuba configure` can update those sections without overwriting any existing files.


template/base/_.gitignore
```diff
@@ -14,3 +15,5 @@ node_modules*/
 npm-debug.log
 package-lock.json
 yarn-error.log
+# end managed by skuba
+
```
72636c, line 19: We need to be consistent with file endings here and/or reinstate the sanitisation logic. Looks like there's an extra newline here, trailing space in `_.eslintignore`, and no trailing newline in `_.dockerignore`

src/cli/configure/analysis/__snapshots__/project.test.ts.snap
```diff
@@ -64,11 +70,15 @@ node_modules*/
 npm-debug.log
 package-lock.json
 yarn-error.log
+# end managed by skuba
+
+
```
72636c, line 75: This doesn't seem right; do we need to trim excess newlines?

src/cli/configure/processing/ignoreFile.ts
```diff
@@ -1,27 +1,13 @@
...
 ) => {
-  const templateLines = splitIntoSanitisedLines(templateFile);
-
-  const templateSet = new Set(templateLines.filter((line) => line !== ''));
-
-  const inputLines = splitIntoSanitisedLines(inputFile ?? '').filter(
-    (line) => !templateSet.has(line),
+  const replacedSkubaSection = inputFile?.replace(
+    /# managed by skuba[\s\S]*# end managed by skuba/,
+    '',
```
72636c, line 6: Can we do the replacement inline here?

How about inverting this to be more like:

```typescript
if (typeof inputFile === 'undefined') {
  return templateFile;
}

return inputFile.replace('...', templateFile);
```

You could even one-liner it, but we might add some extra hacks in here later to make the migration less jarring.

.changeset/pretty-dragons-know.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+Change `configure` to update skuba managed sections in ignore files
```
72636c, line 5: Just a consistency thing

```suggestion
**configure:** Reserve skuba-managed sections in ignore files
```

src/cli/configure/analysis/__snapshots__/project.test.ts.snap
lumishrestha: I think this is from `.gitignore` having an extra new line in the template

src/cli/configure/processing/ignoreFile.ts
```diff
@@ -1,25 +1,20 @@
...
+    const replacedSkubaSection = inputFile.replace(
+      /# managed by skuba[\s\S]*# end managed by skuba/,
+      templateFile,
+    );
 
-  const inputLines = splitIntoSanitisedLines(inputFile ?? '').filter(
-    (line) => !templateSet.has(line),
-  );
+    return `${replacedSkubaSection}`;
+  }
```
72636c, line 15: What do you think about:

```typescript
const replacedFile = inputFile.replace(/.../, templateFile);

if (replacedFile.includes(templateFile)) {
  return replacedFile;
}
```

It's probably overly-paranoid, but it avoids mismatches between the `.replace` pattern and `.includes` string

PR #66, opened 2020-06-18
Store project type to tailor configuration

72636c:
This introduces a new `package.json#/skuba/type` field that is used by `skuba configure` to tailor logic based on whether we're dealing with an application or package. This could be extended in future to somehow enable a monorepo type, but it's probably too early to say.

### DIFF
--- a/src/cli/configure/modules/package.test.ts
+++ b/src/cli/configure/modules/package.test.ts
@@ -1,5 +1,10 @@
 import { parsePackage } from '../processing/package';
-import { assertDefined, defaultOpts, executeModule } from '../testing/module';
+import {
+  assertDefined,
+  defaultOpts,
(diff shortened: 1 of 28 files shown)


src/cli/configure/modules/package.test.ts
```diff
@@ -79,11 +134,79 @@ describe('packageModule', () => {
     const outputData = parsePackage(outputFiles['package.json']);
 
     assertDefined(outputData);
-    expect(outputData.license).toBe('UNLICENSED');
+    expect(outputData.license).toBe('MIT');
```
etaoins, line 137: :+1:

72636c: 🏎️

PR #69, opened 2020-06-18
Extract ESLint docs

72636c:
Spoiler: we'll force a migration to ESLint in `seek-module-toolkit` so that the path from there to `skuba` is smoother.

I wish we had more documentation around the move but I'm hoping this can be community-contributed. I can't make much up off the top of my head.

### DIFF
--- a/docs/eslint.md
+++ b/docs/eslint.md
@@ -1,0 +1,50 @@
+# ESLint
+
+[ESLint] is _the_ JavaScript linter.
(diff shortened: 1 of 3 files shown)


etaoins:
We can probably feed back the questions from `#typescriptifcation` here. Even though I've spent about a month of my waking life fixing ESLint nothing springs to mind here either.

72636c:
👍 I'll also put out a call for feedback once the `seek-module-toolkit` change is in.

PR #75, opened 2020-06-18
Add more package goodies

72636c:
- Say hello to TSDoc
- Drop module aliasing from `tsconfig.json` for package templates


.changeset/stupid-apples-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
```
72636c, line 2: Good idea to call out linting changes more explicitly

```suggestion
'skuba': minor
```

PR #81, opened 2020-06-19
Use `eslint-config-skuba`

72636c:
This is suspiciously easy; I may cut a beta release to verify it's actually working as intended.


package.json
```diff
@@ -51,9 +51,7 @@
     "ejs": "^3.1.3",
     "enquirer": "^2.3.5",
     "eslint": "^7.2.0",
-    "eslint-config-seek": "^7.0.1",
-    "eslint-plugin-jest": "^23.13.2",
-    "eslint-plugin-tsdoc": "^0.2.5",
```
72636c, line 56: In particular, ESLint's docs seem to indicate that it shouldn't be possible to transitively bundle plugins 😕: https://eslint.org/docs/developer-guide/shareable-configs#publishing-a-shareable-config

package.json
```diff
@@ -51,9 +51,7 @@
...
+    "eslint-config-skuba": "1.0.0",
```
tobyhei, line 54: Near miss https://github.com/scub45t3v3/eslint-config-scuba-squad

package.json
tobyhei: 🤞

package.json
72636c: hope I don't get sued

package.json
72636c: Seems to work 🧙

PR #86, opened 2020-06-20
Isolate modules internally

72636c:
This is recommended to flatten the curve and for Babel builds. We won't promote this to the default config just yet.


72636c:
This is apparently less great than I thought: https://github.com/microsoft/TypeScript/issues/18232#issuecomment-380104366

PR #90, opened 2020-06-21
Remove `export =` recommendation

72636c:
This is a weird TypeScript special and isn't supported by Babel.


72636c:
I'm going to leave this in draft; I think the better outcome would be to get `koa-cluster` to support named exports

etaoins (approved):

PR #94, opened 2020-06-22
Clean up ignore files during migration

72636c:
This should leave repos migrated from `seek-module-toolkit` and pre-#58 ignore files in a better state instead of littering old patterns everywhere.

### DIFF
--- a/src/cli/configure/processing/ignoreFile.ts
+++ b/src/cli/configure/processing/ignoreFile.ts
@@ -1,21 +1,69 @@
+/**
+ * Patterns that are superseded by skuba's bundled ignore file patterns and are
+ * non-trivial to derive using e.g. `generateSimpleVariants`.
+ */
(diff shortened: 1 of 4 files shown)


src/cli/configure/processing/ignoreFile.ts
```diff
@@ -1,21 +1,69 @@
...
+const LEADING_SLASH = /^\//;
+const TRAILING_SLASH = /\/$/;
+
+/**
+ * Generate simple variants of an ignore pattern for exact matching purposes.
+ *
+ * Note that these patterns are not actually equivalent (e.g. `lib` matches more
+ * than `lib/`) but they generally represent the same _intent_.
+ */
+export const generateSimpleVariants = (patterns: string[]) => {
```
etaoins, line 17: :+1:

PR #98, opened 2020-06-23
Update smt migration TL;DR

72636c:


72636c:
Bundling #97 in here too

PR #108, opened 2020-06-24
Suppress dependency deprecation warnings

72636c:
These are annoying and there's little that users can do about it. Most of these come from e.g. transitive Jest dependencies, and Jest has open issues tracking the migration to supported dependencies.

### DIFF
--- a/src/utils/exec.ts
+++ b/src/utils/exec.ts
@@ -1,3 +1,4 @@
+import stream from 'stream';
 import util from 'util';
 
 import concurrently from 'concurrently';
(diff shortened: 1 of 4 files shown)


src/utils/exec.ts
```diff
@@ -8,6 +9,23 @@ import npmWhich from 'npm-which';
...
+class YarnWarningFilter extends stream.Transform {
+  _transform(
+    chunk: any,
+    _encoding: BufferEncoding,
+    callback: stream.TransformCallback,
+  ) {
+    const str = Buffer.from(chunk).toString();
+
+    // Filter out annoying deprecation warnings that users can do little about
+    if (!str.startsWith('warning skuba >')) {
```
etaoins, line 21: When I read the PR description I was excited to figure out which Yarn argument did this.

PR #123, opened 2020-06-29
Include tsx files in ESLint linting

jahredhope:
The current calls for ESLint for both checking and fixing linting errors include an explicit list of extensions to check: `js` and `ts`.

An appropriately configured skuba project is able to compile tsx files to js. Typically resolving JSX to React.createElement commands. As it's possible to write and use tsx files I'd like to have eslint run over these files.

This change will include tsx in files that are checked and fixed for eslint.

### DIFF
--- a/.changeset/ninety-crabs-tell.md
+++ b/.changeset/ninety-crabs-tell.md
@@ -2,4 +2,4 @@
 'skuba': patch
 ---
 
-Include tsx files in ESLint linting
+**format, lint:** Include tsx files in ESLint linting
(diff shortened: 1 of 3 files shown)


.changeset/ninety-crabs-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+Include tsx files in ESLint linting
```
72636c, line 5: Just for consistency (CONTRIBUTING.md is coming soon™️)

```suggestion
**format, lint:** Include tsx files in ESLint linting
```

PR #126, opened 2020-06-29
Add `Jest.mergePreset` helper function

72636c:
This concatenates array options, unlike Jest's built-in `preset` support which discards the original array elements.

### DIFF
--- a/src/api/jest/index.ts
+++ b/src/api/jest/index.ts
@@ -1,0 +1,27 @@
+import { Config } from '@jest/types';
+
+import jestPreset from '../../../jest-preset';
+import { mergeRaw } from '../../cli/configure/processing/record';
+
+type Props = Pick<
+  Config.InitialOptions,
+  | 'collectCoverage'
+  | 'collectCoverageFrom'
+  | 'collectCoverageOnlyFrom'
+  | 'coveragePathIgnorePatterns'
+  | 'coverageThreshold'
+  | 'globals'
+  | 'globalSetup'
+  | 'globalTeardown'
+  | 'setupFiles'
+  | 'setupFilesAfterEnv'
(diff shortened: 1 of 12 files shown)


etaoins:
I like the idea but I think the name is non-obvious. Would something like `Jest.mergePreset` make sense?

src/api/jest/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  | 'collectCoverageOnlyFrom'
+  | 'coveragePathIgnorePatterns'
+  | 'coverageThreshold'
+  | 'globals'
+  | 'globalSetup'
+  | 'globalTeardown'
+  | 'setupFiles'
+  | 'setupFilesAfterEnv'
+  | 'testPathIgnorePatterns'
+>;
```
benjaminpearson, line 19: @72636c Sorry to comment on an old PR, but I'm trying to understand why only a subset of jest configuration options is "Pick"ed in this `type Props`. Couldn't teams be given access to all Jest options? Ones that we would like to use for instance is https://jestjs.io/docs/configuration#clearmocks-boolean and https://jestjs.io/docs/configuration#resetmocks-boolean 

I can definitely add those as types to the list or ignore the typescript issue in our project, just wondering the background context as to why its limited :)

72636c: Hey @benjaminpearson! The idea was to only expose options that "safely" extend our preset and wouldn't conflict with things we'd want to manage internally now or in future. But given how long this allowlist has grown, I'm 50:50 on whether to keep extending it or just allow the full list through.

Those two look perfectly reasonable (I wonder if they could even be defaults at some point).

72636c: This should be resolved in the next release: #798

PR #127, opened 2020-06-30
Add a note for yarn path

lumishrestha:


docs/migrating-from-seek-skuba.md
```diff
@@ -33,6 +33,8 @@ then run the following commands on each one:
 1. `skuba configure`
 1. `yarn format`
 
+Note: You may need to add Yarn to your PATH for the commands to run.
```
72636c, line 36: Could we add a link to the Yarn doco about `yarn global bin`?

PR #130, opened 2020-07-02
Use connection reuse environment variable in Lambda

lumishrestha:
Replaces `keepAlive` property of https agent

### DIFF
--- a/template/lambda-sqs-worker/serverless.yml
+++ b/template/lambda-sqs-worker/serverless.yml
@@ -59,6 +59,7 @@ functions:
     reservedConcurrency: 20
     timeout: 30
     environment:
+      AWS_NODEJS_CONNECTION_REUSE_ENABLED: 1
       DESTINATION_SNS_TOPIC_ARN: !Ref DestinationTopic
       ENVIRONMENT: ${env:ENVIRONMENT}
       REGION: ${self:provider.region}
--- a/template/lambda-sqs-worker/src/services/aws.ts
+++ b/template/lambda-sqs-worker/src/services/aws.ts
@@ -3,7 +3,6 @@ import { Agent } from 'https';
 import { SNS } from 'aws-sdk';
 
 const agent = new Agent({
-  keepAlive: true,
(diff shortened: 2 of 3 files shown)


template/lambda-sqs-worker/src/services/aws.ts
```diff
@@ -3,7 +3,6 @@ import { Agent } from 'https';
 import { SNS } from 'aws-sdk';
 
 const agent = new Agent({
```
72636c, line 5: I think we can get rid of the custom agent entirely 🙂

tobyhei, line 5: Can we remove the agent altogether now?

Looks like `rejectUnauthorized` defaults to true anyway. https://nodejs.org/api/tls.html

template/lambda-sqs-worker/serverless.yml
```diff
@@ -59,6 +59,7 @@ functions:
     reservedConcurrency: 20
     timeout: 30
     environment:
+      AWS_NODEJS_CONNECTION_REUSE_ENABLED: 1
```
tobyhei, line 62: I'm on the fence as to whether it's worth even mentioning but in a similar vein but in a much less important way we could set `AWS_STS_REGIONAL_ENDPOINTS=regional` https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html

72636c: Ooh awesome! It would be nice to include this and also have links to the relevant doco for both environment variables, as part of documenting more of the rationale behind these templates

lumishrestha: @tobyhei any reasoning behind not adding that?

tobyhei: Oh sorry, I was just going to suggest we add  `AWS_STS_REGIONAL_ENDPOINTS=regional` but then thought it was scope creep for this PR so removed the comment

PR #150, opened 2020-07-16
Upgrade node: 12 to 14

runk:
Not sure if it's too aggressive but node 14 been around for some time now.

Leaving node 12 only for lambda worker - latest version supported by AWS.


runk:
I assume it's not a breaking change, mark it as minor?

etaoins:
We usually track Node LTS (12) instead of Node stable (14). I think tracking 12 would be too aggressive but I'd be happy to *test* against 14 in addition to 12.

runk:
Did you mean "tracking 14"?

What's the reason behind it? It's stable, right?

etaoins:
Node.js recommends 12 themselves (in addition to the Lambda issue):
Screen Shot 2020-07-16 at 11 20 30

runk:
I know I know, however 14 will be new LTS after some time right, which kinda says it's safe to use (oppose to odd version number). 🤷‍♂️ 

So what's the conclusion? Okay for CI & templates (except lambda) and NOT okay for skuba itself?

tobyhei:
It seems fine for individual teams to opt into newer versions but current LTS accross the board makes sense to me as a default.

Further advice from Node seems to suggest we stick with 12 until 14 is active LTS in October:

> Production applications should only use Active LTS or Maintenance LTS releases.
from https://nodejs.org/en/about/releases/

etaoins:
Yup, I'm fine supporting & testing Node 14 but I don't think it's a sensible default until it become LTS

runk:
Alright, given this let's part this pr till then

PR #151, opened 2020-07-16
Disable yup input coercion

72636c:
This may be the final straw. Should I just switch this over to Runtypes + `runtypes-filter`?

### DIFF
--- a/.changeset/rare-ducks-leave.md
+++ b/.changeset/rare-ducks-leave.md
@@ -1,0 +1,21 @@
+---
+'skuba': patch
+---
+
+**template/koa-rest-api:** Disable yup input coercion
+
+By default, `yup` is ridiculously permissive in coercing inputs to match a schema:
+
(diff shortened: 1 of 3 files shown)


.changeset/rare-ducks-leave.md
```diff
@@ -0,0 +1,21 @@
...
+```typescript
+const birb = yup.object().shape({
+  name: yup.string().notRequired(),
+});
+
+birb.validateSync({ name: 123 });
+// { name: '123' }
+
+birb.validateSync({ name: [{ areYouKiddingMe: no }] });
+// { name: '[object Object]' }
```
etaoins, line 18: This is disappointing

72636c:
Sorry, not doing this: https://github.com/jquense/yup/issues/670#issuecomment-577338523

Closing in favour of a Runtypes replacement

PR #163, opened 2020-07-24
feat: Simplify supertest-koa bootstrap

runk:
Don't start the server and avoid the need to stop it.

> app.callback()
> Return a callback function suitable for the http.createServer() method to handle a request. You may also use this callback function to mount your Koa app in a Connect/Express app.

### DIFF
--- a/template/koa-rest-api/src/testing/server.ts
+++ b/template/koa-rest-api/src/testing/server.ts
@@ -1,5 +1,3 @@
-import { Server } from 'http';
-
 import Router from '@koa/router';
 import Koa, { Middleware } from 'koa';
 import request from 'supertest';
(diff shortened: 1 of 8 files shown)


template/koa-rest-api/src/testing/server.ts
```diff
@@ -10,19 +10,8 @@ import { createApp } from 'src/framework/server';
...
-  const setup = async () => {
-    await new Promise((resolve) => (server = app.listen(undefined, resolve)));
-    agent = request.agent(server);
-  };
-
-  const teardown = () => new Promise((resolve) => server.close(resolve));
-
-  return Object.assign(getAgent, { setup, teardown });
+  const agent: request.SuperTest<request.Test> = request.agent(app.callback());
+  return () => agent;
```
72636c, line 14: I wonder if we should just return `agent` directly from all these helpers? That way we can just do `agent.do()` rather than `agent().do()`

runk: 👍

PR #165, opened 2020-07-26
fix(koa-rest-api): Use per Gantry environment concurrency group in dev

tobyhei:
Matches the production concurrency group configuration.


tobyhei:
I believe we should use this now Gantry uses it's own lock anyway we can't get around it.

PR #176, opened 2020-08-18
Express template

mieshalannair:
Introducing a simple `Hello World` expressjs template for skuba.


README.md
```diff
@@ -143,6 +143,10 @@ This initialises a new directory and Git repository.
 
   A REST [resource API] built on the [Koa] web framework and deployed with [Gantry].
 
+  - `express-rest-api`
+
+  A Plain Express framework and deployed with [Gantry].
```
72636c, line 148:
```suggestion
  A REST [resource API] built on the [Express] web framework and deployed with [Gantry].
```

README.md
```diff
@@ -179,6 +183,7 @@ This initialises a new directory and Git repository.
 [semantic-release]: https://github.com/semantic-release/semantic-release/
 [serverless]: https://serverless.com/
 [worker]: https://tech-strategy.ssod.skinfra.xyz/docs/v1/components.html#worker
+[expressjs]: https://expressjs.com/
```
72636c, line 186:
```suggestion
[express]: https://expressjs.com/
```

src/cli/init/prompts.ts
```diff
@@ -76,6 +76,7 @@ export const TEMPLATE_PROMPT = new Select({
   choices: [
     'greeter',
     'koa-rest-api',
+    'express-rest-api',
```
72636c, line 79: Would you mind alphabetically sorting this here, in the README, etc? It means you get top billing 🙂

template/express-rest-api/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+COPY --from=deps /workdir/node_modules node_modules
+
+ENV NODE_ENV production
+
+ARG PORT=8001
+ENV PORT ${PORT}
+EXPOSE ${PORT}
+
+CMD npx --no-install koa-cluster lib/app
```
72636c, line 67: I wouldn't expect this to work with Express. Have you been able to test deploying a barebones service to a development environment using this template?

template/express-rest-api/src/main.ts
```diff
@@ -0,0 +1,18 @@
+import { Application, Request, Response } from 'express';
+import express = require('express');
```
72636c, line 2: Why does this use a `require`?

template/express-rest-api/src/main.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const app: Application = express();
+
+app.get('/', function (_req: Request, res: Response) {
```
72636c, line 6: Can we use arrow functions here?

template/express-rest-api/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "lint": "skuba lint",
+    "start": "skuba start",
+    "test": "skuba test"
+  },
+  "skuba": {
+    "entryPoint": "src/main.ts",
+    "template": "koa-rest-api",
+    "type": "application",
+    "version": "3.7.7"
+  }
```
72636c, line 31: This stuff is generated

```suggestion
  }
```

template/express-rest-api/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "express": "^4.17.1",
+    "skuba-dive": "^1.1.0"
+  },
+  "devDependencies": {
+    "@types/express": "^4.17.7",
+    "@types/node": "^14.0.27",
+    "@types/supertest": "^2.0.10",
+    "skuba": "^3.7.7",
+    "supertest": "^4.0.2",
+    "typescript": "^3.9.7"
```
72636c, line 12: TypeScript is bundled with skuba, and the version gets substituted in on init.

```suggestion
    "skuba": "*",
    "supertest": "^4.0.2"
```

mieshalannair: Noted, removed it from the dependencies.
How about the skuba should i change that from `"skuba": "^3.7.7"` => `"skuba": "*"`

template/express-rest-api/package.json
mieshalannair: Sorry, if i get you correctly. Should I remove this line ? 😕 
 How about the entry point ?

72636c: The whole `skuba` section can be removed, you'll notice that none of the other templates have it. It's generated dynamically in part from the `skuba.template.js`, e.g.

https://github.com/mieshalannair/skuba/blob/express-template/template/express-rest-api/skuba.template.js#L6

I believe that specified entry point is mismatched with your template. Is there a reason to deviate and choose `main.ts` when every other template uses `app.ts`?

template/express-rest-api/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+COPY --from=deps /workdir/node_modules node_modules
+
+ENV NODE_ENV production
+
+ARG PORT=3000
+ENV PORT ${PORT}
+EXPOSE ${PORT}
+
+CMD node lib/main
```
72636c, line 67: Is this what you run in production? Granted I wouldn't expect these frameworks to properly crash often, but spinning up a new ECS task would be slower than having a clustering implementation spawn another process.

template/express-rest-api/package.json
mieshalannair: Noted with thanks. Will make the changes.
`main.ts` is usually used in the nestjs framework, but I think `app.ts` will do.

template/express-rest-api/Dockerfile
mieshalannair: Yes, we are running our some of our project like this in production. Do you recommend running node with [pm2-runtime](https://pm2.keymetrics.io/docs/usage/docker-pm2-nodejs/) ?

72636c: I think there's an open question for us to resolve around when multiple processes make sense with constrained compute in Fargate. I'm happy to leave this for now.

aleksei-seekasia: Hi @72636c , we did use pm2 for one of our services in the past, but stopped. Main reason: we don't expect crashes to be a part of a normal application lifecycle. It also was causing us some issues, when app was keep crashing but we thought it was running (but probably because of the way, how we configured our monitors in DataDog).

> I think there's an open question for us to resolve around when multiple processes make sense with constrained compute in Fargate.

I see this question is more like an advanced topic, which requires to have proper circumstances. We don't have it yet. So I would agree with you that we might leave it out of scope for now.

p.s. on a personal note I would like to hear when and how you use pm2 to overcome "constrained compute in Fargate".

72636c (approved):
🚢 thanks for this!

PR #181, opened 2020-08-23
Retain package comments on first configure

72636c:
A number of our older packages don't seem to retain comments in the compiled TypeScript. This makes the first configure (i.e. when migrating from another toolkit) default to retaining comments.


etaoins (approved):
:+1:

PR #186, opened 2020-08-26
Qualify `awsRequestId` log field

72636c:
Plain `requestId` was a bit ambiguous as we also have concepts like an `X-Request-Id` header.


etaoins (approved):

tobyhei (approved):
🥳

PR #200, opened 2020-09-06
Support `tsc --build` flag

72636c:
This is a small step towards better monorepo support.

https://www.typescriptlang.org/docs/handbook/project-references.html


etaoins (approved):
😮

PR #209, opened 2020-09-22
feat(rest api): Add default CPU and memory settings

tobyhei:
Adds lower default memory and CPU values for Koa apps.

As @etaoins mentioned elsewhere, the higher Gantry defaults make sense without knowing what the image/runtime will be, e.g a JVM app that may need more memory just to start up.

For skuba where we know the runtime is Node, the lowest available Fargate options seem like a sane default.

### DIFF
--- a/rich-books-confess.md
+++ b/rich-books-confess.md
@@ -1,0 +1,5 @@
+---
+'skuba': patch
+---
+
+**template/*-rest-api:** Reduce default CPU and memory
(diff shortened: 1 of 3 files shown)


etaoins (approved):
:+1:

This is what we use for nearly all of our prod services.

72636c (approved):
Ta, can you do the same for the Express template?

PR #221, opened 2020-09-28
Switch default branch to `main`

72636c:
Our last September deadline is GitHub's new default branch. Existing repos are unaffected, so this is concerned with initialising new repos with `main`.


72636c:
semantic-release doesn't seem to be ready for this, so we may need to override its config.

72636c:
Not seeing a lot of buy in from internal tooling at this stage. We can revisit this in future.

PR #222, opened 2020-09-28
Skip pre-build in Docker Compose service

72636c:
This should be generally faster across the pipelines, especially for Gantry services that already have this covered with `gantry build`. A caveat is that the Lambda deploy step may take a bit longer, but it's probably good to have a safer npm `deploy` script and we know that we need to revisit the speed of that pipeline soon anyway.


etaoins (approved):
:+1:

PR #225, opened 2020-10-01
Adopt `@seek/logger`

72636c:
https://github.com/seek-oss/logger


kosanna (approved):
Yep cool, looks good, just need to hold on until we transfer

tobyhei (approved):
🎉

Issue #282, opened 2020-12-03
README links to the Gantry repo, which is not viewable by most people


kriskhaira:
Hi. The README currently says "Deploy with Gantry" but not everyone can view the Gantry repo. This can create some [confusion](https://seek-asia.slack.com/archives/C21MWHLDT/p1606111050218300?thread_ts=1606106039.212900&cid=C21MWHLDT) for the general public (since skuba is an OSS), and for SEEK Asia engineers (since not all of us can view SEEK-Jobs repos).

PR #287, opened 2020-12-04
Use official doc URL instead of repo for Gantry

kriskhaira:
This makes it less confusing for SEEK Asia engineers who are not able to view seek-jobs/gantry, and addresses #282.


72636c (approved):
🙇

PR #289, opened 2020-12-07
Fix server listener and port in Express template

72636c:
The template wasn't respecting the PORT environment variable provided by a Gantry deployment, and wasn't compatible with `skuba start`.

This fixes those issues and introduces some boilerplate config and logging code that can be extended in future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,7 @@ As does Express:
 ```typescript
 const app = express();
 
-export default app;
+export default Object.assign(app, { port });
 ```
 
 ### `skuba test`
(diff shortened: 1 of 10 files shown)


tobyhei (approved):

README.md
```diff
@@ -262,6 +262,14 @@ const app = new Koa();
 export default Object.assign(app, { port });
 ```
 
+As does Express:
+
+```typescript
+const app = express();
+
+export default app;
```
72636c, line 270:
```suggestion
export default Object.assign(app, { port });
```

PR #290, opened 2020-12-08
Tweak template test scripts

72636c:
I've been umming and ahhing on the right approach here, as I like a fairly unopinionated `test` script into which the user can pass their desired flags.

Practically speaking, we have some other tooling like our package publishing pipeline which presumes that the `test` script is suitable for exhaustive CI testing. `test:watch` should be a reasonable compromise for local testing, and already has some SEEK precedence with ~46 `package.json` results returned from a GitHub search.

### DIFF
--- a/.changeset/cuddly-pillows-push.md
+++ b/.changeset/cuddly-pillows-push.md
@@ -1,0 +1,5 @@
+---
+'skuba': patch
+---
+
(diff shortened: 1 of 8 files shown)


etaoins (approved):
Seems reasonable

PR #298, opened 2020-12-14
Implement node wrapper

### LINKED ISSUE #84: `skuba exec`
It would be nice to have a simple counterpart to `skuba start` that executes the provided TypeScript source file without starting up a persistent reloading process. This would presumably use `ts-node` for `tsc` projects and `babel-node` for prospective Babel projects.

72636c:
See changesets and README for full details. Closes #84.

---

`skuba node` runs a TypeScript file; it's like a one-off `skuba start`. Unlike start, it gives you a REPL when you omit the command line entry point rather than falling back to the entry point in `package.json`.

This PR also adds some secret sauce to automatically register the `src` module alias when you're running under `skuba node` or `skuba start`. This means that you only need a runtime module alias resolver like `import skuba-dive/register` at the top of production entry points; local scripts can be run with little fanfare using the skuba commands.

---

Notes:

- I considered calling the command `skuba exec`, but that sounds a bit awkward when it supports opening a REPL.

- You may be wondering why don't we just use `tsconfig-paths`.

  This is tempting but I'm not convinced that we should depend on reading configuration out of `tsconfig.json` at runtime. This can be confusing given TypeScript is generally considered a compile-time concern, and our own templates only keep `lib` and `node_modules` in the runtime Docker image.

- Diving into this has uncovered that the `@babel/node` REPL is pretty unusable with modern JavaScript/TypeScript. See the README for more.

### DIFF
--- a/.changeset/late-carrots-speak.md
+++ b/.changeset/late-carrots-speak.md
@@ -1,0 +1,7 @@
+---
+'skuba': patch
+---
+
+**start:** Improve support for non-HTTP server entry points
+
+You can now run arbitrary TypeScript files without them exiting on a `You must export callback or requestListener` error.
--- a/.changeset/light-files-nail.md
+++ b/.changeset/light-files-nail.md
@@ -1,0 +1,33 @@
+---
+'skuba': minor
+---
+
+**node:** Add command
+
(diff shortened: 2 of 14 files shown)


tobyhei (approved):
🤩

etaoins (approved):
👏

PR #300, opened 2020-12-14
Check for newer skuba versions

### LINKED ISSUE #109: Check for new versions on `skuba configure` and `skuba init`
Global installations of `skuba` are much faster to work with than having `npx` reinstall the world each time you want to initialise or configure a repo. The downside is that it would be easy for these installation to fall out-of-date. This could be addressed by adding an optimistic check for a new version of the package using e.g. `latest-version`.

I'd limit this to the listed commands as the core development commands (e.g. `build`) should remain simple, fast and offline.

72636c:
Closes #109.

|offline|up to date|out of date
|:-:|:-:|:-:
Screen Shot 2020-12-14 at 2 58 31 pm | Screen Shot 2020-12-14 at 2 59 09 pm | Screen Shot 2020-12-14 at 2 59 31 pm

### DIFF
--- a/.changeset/thick-dancers-taste.md
+++ b/.changeset/thick-dancers-taste.md
@@ -1,0 +1,7 @@
+---
+'skuba': minor
+---
+
+**configure, help, init:** Check for newer skuba versions
+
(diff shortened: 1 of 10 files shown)


etaoins (approved):
When are you going to start asking for donations?

PR #301, opened 2020-12-14
Support function entry points

### LINKED ISSUE #115: Local execution for Lambdas
The `lambda-sqs-worker` template does not come with a `skuba start` script. This is because we haven't decided on a local execution experience for Lambdas. A few options in my current order of preference:

1. Serve the handler over HTTP so you can POST the input events. I consider this slightly better than option 2 because it's less boilerplate and lets you test container reuse scenarios.

1. Just require the entry point. This will likely mean that users will need to provide their own local wrapper around the Lambda handler, e.g.

   ```typescript
   // localHandler.ts

   import { handler } from './app'

   handler({ event: 'my test event', context: 'my test context' })
   ```

1. A local Lambda runtime, e.g. proxying [`serverless invoke local`](https://www.serverless.com/framework/docs/providers/aws/cli-reference/invoke-local/). I'm not super keen on replicating the runtime locally or tying `skuba` to Serverless but could be convinced otherwise.

72636c:
This enables running a Lambda handler as a HTTP server locally using `skuba node` or `skuba start`. Because HTTP is the best FFI, obviously.

Closes #115.


src/wrapper/functionHandler.ts
```diff
@@ -0,0 +1,38 @@
...
+}: Args) => {
+  if (!isObject(entryPoint)) {
+    log.subtle(log.bold(functionName), 'is not exported');
+    return;
+  }
+
+  const fn = entryPoint[functionName];
+
+  if (!isFunction(fn)) {
+    log.subtle(log.bold(functionName), 'is', typeof fn, 'not a function');
```
72636c, line 31:
```suggestion
    log.subtle(log.bold(functionName), 'is not a function');
```

etaoins (approved):
Still better than APIGW

PR #310, opened 2021-01-03
Use explicit deregistrationDelay in Gantry

tobyhei:
The default of 120 seconds seems to be excessive for most workloads, so provide a more sensible default explicitly.

This will hopefully also prompt users to tweak the value as they see fit.

A typical Gantry deploy seems to take about 5 minutes, so this should reduce typical time to deploy by about a third.

### DIFF
--- a/.changeset/fresh-ladybugs-sip.md
+++ b/.changeset/fresh-ladybugs-sip.md
@@ -1,0 +1,5 @@
+---
+"skuba": patch
+---
+
+Use explicit deregistrationDelay in Gantry
--- a/.changeset/fresh-ladybugs-sip.md
+++ b/.changeset/fresh-ladybugs-sip.md
@@ -1,5 +1,5 @@
 ---
-"skuba": patch
+'skuba': patch
 ---
 
-Use explicit deregistrationDelay in Gantry
+**template:** Use explicit `deregistrationDelay` in Gantry
--- a/template/express-rest-api/gantry.apply.yml
+++ b/template/express-rest-api/gantry.apply.yml
@@ -48,6 +48,8 @@ deployment:
     path: /health
     useExternalDns: true
 
+deregistrationDelay: 10
+
 tags:
   {{range $key, $value := .Values.tags}}
   {{$key}}: '{{$value}}'
--- a/template/koa-rest-api/gantry.apply.yml
+++ b/template/koa-rest-api/gantry.apply.yml
@@ -49,6 +49,8 @@ deployment:
     path: /smoke
     useExternalDns: true
 
+deregistrationDelay: 10
+
 tags:
   {{range $key, $value := .Values.tags}}
   {{$key}}: '{{$value}}'


etaoins:
Does this mean any requests that take longer than 10 seconds will be terminated with a 500 during a deploy? That seems aggressive and I'm not sure developers connect the dots between the 500s and needing to change this setting.

According to the docs deregistration should be skipped completely if there are no active connections (presumably the usual state for `dev` services and a large number of `prod` ones):
> If a deregistering target has no in-flight requests and no active connections, Elastic Load Balancing immediately completes the deregistration process, without waiting for the deregistration delay to elapse

tobyhei:
> According to the docs deregistration should be skipped completely

It looks like that might be a bit misleading https://serverfault.com/questions/919335/alb-connection-draining-is-always-reaching-the-deregistration-delay

> Does this mean any requests that take longer than 10 seconds will be terminated with a 500 during a deploy? That seems aggressive and I'm not sure developers connect the dots between the 500s and needing to change this setting.

Hmm do you have a less aggressive target in mind, or do you think 120 seconds is already appropriate?

etaoins:
> https://serverfault.com/questions/919335/alb-connection-draining-is-always-reaching-the-deregistration-delay
Well that's disappointing

> Hmm do you have a less aggressive target in mind, or do you think 120 seconds is already appropriate?

When I was in Candy we used to have to deal with user's browsers on e.g. mobile phones that might take forever to complete an HTTP transaction. Now that everything is behind a GraphQL it could make more sense to be a bit more aggressive everywhere but the GraphQL server.

However, even in the SEEK API there are things that could take more than 10s to complete. I think if we expect that to complete normally we should also expect it to complete during a deploy.

There's also a question if this belongs in Skuba or Gantry. Where Skuba has deviated from defaults before it was usually somewhere we'd expect a greenfield TypeScript service to deviate from Gantry's wider userbase. I don't think the deregistration delay would vary between implementations; it's more about what the expected timeouts on the HTTP API are.

I don't feel super strongly though. I'd like to hear what @72636c & @bernos think

72636c:
I think it makes sense to leave the default to Gantry, and 10s does seem a bit aggressive. We incur this penalty after traffic migration, which wouldn't really affect our time to deploy (not counting the knock-on effect of the concurrency lock).

How does this interact with the `terminationWaitTimeInMinutes`, which defaults to 1 minute?

tobyhei:
They’re serial afaik

tobyhei:
> We incur this penalty after traffic migration, which wouldn't really affect our time to deploy (not counting the knock-on effect of the concurrency lock).

We block prod on completing the dev deploy too.

tobyhei:
> How does this interact with the terminationWaitTimeInMinutes, which defaults to 1 minute?

Seems this runs concurrently as you expected so the default of 2 minutes isn't actually respected anyway and there's no benefit in reducing the `deregistrationDelay` below or above `terminationWaitTimeInMinutes` anyway. 😕

PR #322, opened 2021-01-20
Supply custom autoscaling policy to Gantry

72636c:
I expect this to be spicy but it seems like a more sensible default for typical Node.js HTTP APIs.


etaoins (approved):
Makes sense for now. I'd love to find a backward compatible way to disable request based autoscaling in Gantry.

tobyhei (approved):
Would be nice for Gantry to have better defaults but this works for me

Issue #116, opened 2020-06-24
Smoke testing in templates


72636c:
- The `lambda-sqs-worker` template doesn't prescribe any pattern around invoking the Lambda function post-deployment as a smoke test, though we tend to do this internally. Rollbacks would be a natural follow-up question to that.

- The `koa-rest-api` template doesn't really explain what the smoke test is supposed to do. We tend to find it most useful to validate that the runtime environment has appropriate permissions to access any dependent resources.

PR #328, opened 2021-01-26
Add smoke testing to Lambda template

72636c:
This is missing a bit of documentation; I have a follow-up PR to uplift the API and worker READMEs. We've tested this pattern with an internal service and I've also ensured that the template is deployable from scratch.

---

Relates to #116, which can be closed once we add appropriate documentation to the API templates.

### DIFF
--- a/template/lambda-sqs-worker/serverless.yml
+++ b/template/lambda-sqs-worker/serverless.yml
@@ -8,6 +8,13 @@ custom:
       isProduction: 'false'
     prod:
       isProduction: 'true'
+  prune:
+    automatic: true
+    number: 3
+
+plugins:
+  - serverless-plugin-canary-deployments
+  - serverless-prune-plugin
 
 provider:
   logRetentionInDays: 30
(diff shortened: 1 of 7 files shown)


etaoins (approved):
🥈

template/lambda-sqs-worker/serverless.yml
```diff
@@ -17,15 +24,25 @@ provider:
...
+  versionFunctions: true
   # deploymentBucket:
   #   name: 'TODO: deploymentBucketName'
   iamRoleStatements:
+    - Action: codedeploy:PutLifecycleEventHookExecutionStatus
+      Effect: Allow
+      Resource: !Join
+        - ''
+        - - !Sub arn:aws:codedeploy:${AWS::Region}:${AWS::AccountId}:deploymentgroup
+          - ':*/'
```
72636c, line 36: The `*` is not ideal but the logical ID for the CodeDeploy application is difficult to predict without access to plugin internals. I've PRed `serverless-plugin-canary-deployments`  to automatically infer this statement, but I don't want to block our next release on it being accepted.

PR #332, opened 2021-01-26
Clarify API health checks and smoke tests

### LINKED ISSUE #116: Smoke testing in templates

72636c:
Resolves #116.

### DIFF
--- a/template/koa-rest-api/src/app.ts
+++ b/template/koa-rest-api/src/app.ts
@@ -6,6 +6,6 @@ import { createApp } from './framework/server';
 
 const app = createApp(router.allowedMethods(), router.routes());
 
-Object.assign(app, { port: config.port });
-
-export default app;
+export default Object.assign(app, {
+  port: config.port,
+});
--- a/template/koa-rest-api/src/app.ts
+++ b/template/koa-rest-api/src/app.ts
@@ -1,9 +1,11 @@
-import 'skuba-dive/register';
+import './register';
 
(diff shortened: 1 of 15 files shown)


template/koa-rest-api/src/app.ts
```diff
@@ -1,9 +1,11 @@
...
+import { router } from './api';
+import { config } from './config';
+import { createApp } from './framework/server';
 
 const app = createApp(router.allowedMethods(), router.routes());
 
-export default Object.assign(app, { port: config.port });
+Object.assign(app, { port: config.port });
+
+export default app;
```
72636c, line 11: Inconsequential change, but I tried to make it a bit more obvious that `Object.assign` is mutating here.

template/koa-rest-api/src/listen.ts
```diff
@@ -1,3 +1,5 @@
+import './register';
+
```
72636c, line 2: Making this explicit now because you can run into issues if `./app` is not imported first.

tobyhei (approved):

PR #333, opened 2021-01-26
Pick random server port on init

72636c:
This gives the initialised template a stable port to `skuba start` on. With any luck the randomness will also reduce the chance of collisions in the event that you need to run multiple projects locally at once.


etaoins (approved):
Nice solution :+1:

PR #343, opened 2021-02-02
Run processes serially on Buildkite

72636c:
This is sad but it's not uncommon to run into issues with typical instance sizing at SEEK. This should also alleviate contention on the shared publishing agents.

If we want to be clever in future we could inspect `BUILDKITE_AGENT_META_DATA_AWS_INSTANCE_TYPE`.

### DIFF
--- a/.changeset/purple-points-wash.md
+++ b/.changeset/purple-points-wash.md
@@ -1,0 +1,7 @@
+---
+'skuba': patch
+---
+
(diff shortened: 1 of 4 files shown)


tobyhei (approved):

72636c:
Longer term I'd like to explore running these in-process rather than spawning child processes. Unfortunately these tools generally don't provide programmatic APIs that match the convenience of their CLIs.

PR #347, opened 2021-02-04
Upgrade templates to Node 14

etaoins:
Node.js 14 is [now supported on AWS Lambda](https://aws.amazon.com/about-aws/whats-new/2021/02/aws-lambda-now-supports-node-js-14/). This lets us upgrade the Node.js requirement for skuba's templates.

This should only impact newly created projects. You can use the template changes in this PR as an example of how to upgrade an existing project.


tobyhei (approved):

tobyhei:
Dupe of https://github.com/seek-oss/skuba/pull/150 ?

etaoins:
@tobyhei 

Good catch. I think also this covers some of the newer templates, though.

template/oss-npm-package/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "outDir": "lib",
     "removeComments": false,
-    "target": "ES2019"
+    "target": "es2020"
```
72636c, line 5: 🎉

package.json
```diff
@@ -5,7 +5,7 @@
     "skuba": "lib/skuba.js"
   },
   "engines": {
-    "node": ">=12"
+    "node": ">=14"
```
72636c, line 8: Yeah _this_ is a major. Do you feel it's worthwhile to force people to upgrade at this time?

.changeset/tender-mugs-cover.md
```diff
@@ -0,0 +1,10 @@
+---
+'skuba': minor
+---
+
+**node**: Upgrade to Node 14
+
+Node 14 is now supported on AWS Lambda:
+<https://aws.amazon.com/about-aws/whats-new/2021/02/aws-lambda-now-supports-node-js-14/>
+
+This lets us upgrade the Node requirement across Skuba & its templates.
```
72636c: GitHub releases infuriatingly disrespect Markdown newline rules:

```suggestion
**all**: Upgrade to Node.js 14

Node.js 14 is [now supported on AWS Lambda](https://aws.amazon.com/about-aws/whats-new/2021/02/aws-lambda-now-supports-node-js-14/). This lets us upgrade the Node.js requirement across skuba & its templates.
```

72636c (approved):
It would be neat if we could figure out some `configure` magic for this too. I might have a look at that separately if I have some spare time.

package.json
etaoins: Yeah, let me have another run at this.

etaoins:
@72636c 

Yeah, I was thinking about that. If we go with the more gentle approach I just pushed we should be able to go out with this and then add `configure` support afterwards.

72636c (approved):
🙇

PR #356, opened 2021-02-15
Retire REGION in favour of AWS_REGION

dan753722:
Why:
1. `AWS_REGION` is the reserved runtime environment variable in AWS lambda.
2. AWS_SDK looks for AWS_REGION by default
3. one less environment configuration in serverless.yml

References:
https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html#setting-region-environment-variable
https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html

### DIFF
--- a/.changeset/silly-knives-prove.md
+++ b/.changeset/silly-knives-prove.md
@@ -1,0 +1,5 @@
+---
+"skuba": patch
+---
+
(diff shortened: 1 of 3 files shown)


72636c (approved):
🙇

PR #363, opened 2021-02-22
Drop region parameterisation

72636c:
🙊

### DIFF
--- a/.changeset/good-eggs-jog.md
+++ b/.changeset/good-eggs-jog.md
@@ -1,0 +1,5 @@
+---
+'skuba': patch
+---
+
+**template:** Drop region parameterisation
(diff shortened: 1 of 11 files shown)


etaoins (commented):
Is this just for simplification? Would non-Sydney and multi-region services be expected to re-add what they need?

There will be fewer of those than we thought but there are also already Skuba services in SEEK Asia. I want to make sure we’re not being too aggressive here.

72636c:
@etaoins yep it's purely simplification for newly initialised repos going forwards, and it would need to be re-added for multi-region services.

Granted this is upstreamed from a much-simplified template I'm building for the grad bootcamp, so maybe it's too aggressive.

etaoins (approved):

tobyhei (approved):
Is there any newly minted tech guidance written down about the default region we could link to yet?

etaoins:
@tobyhei 

All the recent changes are reflected only in Confluence so far. There’s plans to merge things in to Tech Strategy once we have a better handle on what it should be.

For now `ap-southeast-2` is a safe bet, though.

72636c:
Also note that the only tangible difference here is that a newly initialised service will not log its `region`. That feels like a reasonable default and saves some $$$ since most of our services are not multi-region.

PR #373, opened 2021-03-01
Fail Gantry build if ECR scanning reports vulnerabilities

renkoh:
```
failOnScanFindings (optional / boolean)
Set this property to true to fail build when ECR scanning reports vulnerabilities. Defaults to false.
```

Set this to true for all new projects which use Gantry to build/deploy. This should help to ensure that we don't use vulnerable containers or introduce vulnerable binaries into our containers.


template/express-rest-api/gantry.build.yml
```diff
@@ -6,3 +6,5 @@ image: '{{values "image"}}'
 
 buildArgs:
   NPM_READ_TOKEN: '{{.Env.NPM_READ_TOKEN}}'
+
+failOnScanFindings: true
```
72636c, line 10: I find your lack of final newlines disturbing

72636c (approved):
🙇

PR #374, opened 2021-03-02
Move Gantry region config to plugin options

72636c:
The next version of the Gantry plugin will start requiring this option.


bridgetbarnes (approved):
I want to be as proactive as you are someday

tobyhei (approved):

raphsutti (approved):

PR #389, opened 2021-03-25
Expose testTimeout in jest preset options

tobyhei:
Makes it easier to set without having to merge with a separate object, e.g https://github.com/SEEK-Jobs/indie-awsk-redux/pull/23 

~I'm not sure if there is a way to set this if it isn't configurable via the skuba preset.~

### DIFF
--- a/.changeset/perfect-eyes-approve.md
+++ b/.changeset/perfect-eyes-approve.md
@@ -1,0 +1,5 @@
+---
+"skuba": patch
+---
+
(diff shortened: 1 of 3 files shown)


tobyhei:
nvm, I misunderstood what was going on here

72636c:
The `Pick` is a bit restrictive and tries to limit things to the props that you'd reasonably want to tweak, and I think this is one of them.

72636c (approved):

PR #418, opened 2021-04-16
Add commitizen to package templates

kosanna:
We use semantic release as a primary release mechanism for private packages, and many teams and individuals still struggle from not being able to construct semantic commit names, so adding `commit` script to assist with on-boarding.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "babel-plugin-macros": "^3.0.1",
     "babel-plugin-module-resolver": "^4.1.0",
     "chalk": "^4.1.0",
-    "commitizen": "^4.2.3",
     "concurrently": "^6.0.0",
     "ejs": "^3.1.6",
     "enquirer": "^2.3.6",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "express": "4.17.1",
     "koa": "2.13.1",
     "supertest": "6.1.3",
-    "type-fest": "1.0.1"
+    "type-fest": "1.0.2"
   },
   "dependencies": {
     "@babel/cli": "^7.12.10",
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "babel-plugin-macros": "^3.0.1",
     "babel-plugin-module-resolver": "^4.1.0",
     "chalk": "^4.1.0",
+    "commitizen": "^4.2.3",
     "concurrently": "^6.0.0",
     "ejs": "^3.1.6",
     "enquirer": "^2.3.6",
--- a/template/private-npm-package/README.md
+++ b/template/private-npm-package/README.md
@@ -28,6 +28,7 @@ Next steps:
   - [Lint](#lint)
   - [Package](#package)
 - [Release](#release)
+  - [Commit messages](#commit-messages)
   - [Releasing latest](#releasing-latest)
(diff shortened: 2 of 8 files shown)


template/private-npm-package/_package.json
```diff
@@ -21,11 +21,17 @@
   "scripts": {
     "build": "skuba build-package",
     "format": "skuba format",
+    "commit": "cz",
```
72636c, line 24:
```suggestion
    "commit": "cz",
    "format": "skuba format",
```

package.json
```diff
@@ -62,6 +62,7 @@
     "babel-plugin-macros": "^3.0.1",
     "babel-plugin-module-resolver": "^4.1.0",
     "chalk": "^4.1.0",
+    "commitizen": "^4.2.3",
```
72636c, line 65: I'm not keen on bundling commitizen into skuba itself while it has [high-severity security vulnerabilities](https://snyk.io/test/npm/commitizen/4.2.3) that have gone unpatched for months. If we're sure this is still the best tool for the job (what does it do besides present ~2 prompts to the user?) I'd place it in the template `package.json`.

template/private-npm-package/_package.json
```diff
@@ -21,11 +21,17 @@
...
     "lint": "skuba lint",
     "release": "yarn build && skuba release",
     "test": "skuba test --coverage",
     "test:watch": "skuba test --watch"
   },
+  "config": {
```
72636c, line 30: 1. I believe we can/should do the same to the `oss-npm-package` template.
2. Should we add a section to the relevant template READMEs?

kosanna: Ok, I was just not sure about the `oss`, just because you could go either way about it and implement `changesets` for example

72636c: 👍 It's definitely easier to get Changesets going in `oss`, but it should be fine to include this out of the box since the template defaults to using Semantic Release anyway.

template/private-npm-package/README.md
```diff
@@ -120,3 +120,7 @@ For more information, see the **semantic-release** docs on [triggering a release
 [seek-jobs/renovate]: https://github.com/SEEK-Jobs/renovate
 [semantic-release]: https://github.com/semantic-release/semantic-release
 [triggering a release]: https://github.com/semantic-release/semantic-release/#triggering-a-release
+
+In order to facilitate the alignment with semantic release commit naming convention you can use `yarn commit` command. The script will take you through a range of questions, in the end of which you will have a correct commit name and description for semantic release to pick it up.
+
+However, if you are using _squash and merge_ commits you will need to ensure the convention is preserved upon merging the pull request on Github by ensuring title and description are still populated with necesary details and follow the convention.
```
72636c, line 126: Should this go in its own section under `Release`? Something like:

```markdown
## Release

...

### Commit messages

Here!

### Releasing latest

...
```

package.json
kosanna: Cool, can move it to templates then 👍

kosanna: I guess sec is less of a concern for dev deps on packages

template/private-npm-package/README.md
kosanna: As you wish, I think it makes sense 👌

72636c (approved):
Thanks for this!

PR #428, opened 2021-05-22
Build lambda-sqs-worker-cdk before deploy

damienungseek:
**Why?**
Prior to this change, the built code was packaged with the docker image. Since the image is ECR cached on `package.json` and `yarn.lock`, stale code will be deployed if there are no changes to either file.

**What?**
This change adds a build command before CDK deploy to ensure the latest built code gets picked up and deployed. It also removes the unnecessary code build in the image.

### DIFF
--- a/template/lambda-sqs-worker-cdk/Dockerfile
+++ b/template/lambda-sqs-worker-cdk/Dockerfile
@@ -15,16 +15,8 @@ RUN \
 
 ###
 
-FROM node:14-alpine AS dev-deps
+FROM node:14-alpine AS build
 
 WORKDIR /workdir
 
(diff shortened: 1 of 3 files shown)


template/lambda-sqs-worker-cdk/Dockerfile
```diff
@@ -15,16 +15,8 @@ RUN \
 
 ###
 
-FROM node:14-alpine AS dev-deps
+FROM node:14-alpine AS build
```
72636c, line 18:
```suggestion
FROM node:14-alpine AS dev-deps
```

72636c (approved):
Nice one, thanks

PR #429, opened 2021-05-27
Banish `typeof undefined` syntax

72636c:
Neither skuba nor the affected templates here would be run in an ES3 environment.


etaoins (approved):
😌

PR #437, opened 2021-06-02
Removing `sqs:SendMessage` permission to Dead Letter Queue for `lambda-sqs-worker`

relientm96:
### Description
- In the `lambda-sqs-worker` template,
- There is no longer a need to specifically grant `sqs:SendMessage*` for your lambda to your DeadLetterQueue
- This is now automatically done by AWS under the hood
- Therefore, we can remove this from the `provider.iam` yaml config
- As long as the `RedrivePolicy` is configured in the `SQS Message Queue`

### Validation

- The `apply-services` team has a new worker that has this removed.
- And DLQ messages are still being sent to the DLQ.
- Link to serverless: https://github.com/SEEK-Jobs/apply-pipeline-job-enricher/blob/629a1e338edef91bbbbab552285c21308b347503/serverless.yml#L25
- Another example of the `virus-scanner-service-submitter` worker also having this removed:
(https://github.com/SEEK-Jobs/virus-scanner-service/blob/379ba1fdbba4db3ddc6d8f72cd59d54014ac9c83/serverless.yml#L27)

### DIFF
--- a/template/lambda-sqs-worker/serverless.yml
+++ b/template/lambda-sqs-worker/serverless.yml
@@ -45,9 +45,6 @@ provider:
         - Action: sns:Publish
           Effect: Allow
           Resource: !Ref DestinationTopic
-        - Action: sqs:SendMessage*
-          Effect: Allow
-          Resource: !GetAtt DeadLetterQueue.Arn
   stackTags:
     # TODO: add data classification tags
     # https://rfc.skinfra.xyz/RFC019-AWS-Tagging-Standard.html#seekdataconsumers


72636c:
While the bundled Lambda templates don't demonstrate this, it's relatively common for productionised Lambda functions to re-queue messages to the input and dead letter queues directly in code rather than leaving it up to invocation failures and redrive policies. In particular, AWS will throttle a failing function which can lead to further degradation.

tobyhei:
Tech strategy on the above https://github.com/SEEK-Jobs/tech-strategy/pull/178/files#diff-3170c20a3b7f0649b0889fc63f22634e8c363dde973931d43d3638363425a2ceR196

tobyhei:
Tech strategy on the above 
https://github.com/SEEK-Jobs/tech-strategy/blob/master/docs/v1/technology.md#performance-degradation-on-lambda-function-errors

```
A Lambda function with high throughput requirements should handle and retry errors in code rather than bubbling them up to the Lambda service as function errors.
```

relientm96:
Thank you for the clarification @72636c and @tobyhei !

Will close this PR 👍

PR #438, opened 2021-06-02
Handcraft a Renovate config

72636c:
skuba has pretty intense requirements for renovation; most notably, it bundles templates that need to be tagged and upgraded using different rules. We've outlasted the usefulness of inherited defaults and our frankenstein config stopped working entirely a month ago, per #426.

This tries to vastly simplify a copy of the old base config.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,8 +1,9 @@
 {
   extends: [
-    ':prHourlyLimit2',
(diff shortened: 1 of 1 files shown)


tobyhei (approved):

72636c:
This will probably end poorly, we'll see what happens in the morning 🙈

PR #447, opened 2021-06-07
Fix terrible changeset

72636c:


tobyhei (approved):
So many new TS features to take advantage of now

PR #453, opened 2021-06-10
Improve ECR caching in templates

72636c:
Editorialised context from an internal PR I forgot to upstream:

---

Our previous attempt at `COMPOSE_DOCKER_CLI_BUILD` works locally for `docker build` and `docker-compose build` interop. The issue in Buildkite is that we use multi-stage builds which means that the cached ECR image only contains layers from the final stage, so Compose will always try to rebuild the earlier stages like `yarn install`.

This switches our Compose file over to directly reference the cached ECR image.

1. Upside: we never wastefully rebuild base images through Compose, saving ~2 minutes per step.
2. Downside: Compose file is not usable for local debugging. A workaround is to define separate Compose services for CI and local use, but that isn't implemented in this PR.


tobyhei (approved):
Definitely worth it for the speed up on long builds IMO 🙇

etaoins (approved):

PR #488, opened 2021-08-05
Parameterise region in Gantry templates

bridgetbarnes:
The region values for Gantry are currently hardcoded to ap-southeast-2, but I'd like to encourage all APAC teams to use skuba so parameterising the region would be nice. 

I guessed this is a patch change based on https://github.com/seek-oss/skuba/commit/985ff9a73cf079882efe7f81e45679e8e0a1f7d1

### DIFF
--- a/template/koa-rest-api/.gantry/common.yml
+++ b/template/koa-rest-api/.gantry/common.yml
@@ -1,7 +1,8 @@
 prodAccountId: '<%- prodAwsAccountId %>'
 
-image: '{{values "prodAccountId"}}.dkr.ecr.ap-southeast-2.amazonaws.com/{{values "service"}}:{{.BuildID}}'
+image: '{{values "prodAccountId"}}.dkr.ecr.{{values "region"}}.amazonaws.com/{{values "service"}}:{{.BuildID}}'
 service: '<%- serviceName %>'
+region: '<%- prodAwsAccountId %>'
 
 # TODO: enable Datadog agent
 # https://gantry.ssod.skinfra.xyz/docs/v1/resources/service.html#datadogSecretId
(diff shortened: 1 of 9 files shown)


template/koa-rest-api/.gantry/common.yml
```diff
@@ -1,11 +1,12 @@
 prodAccountId: '<%- prodAwsAccountId %>'
 
-image: '{{values "prodAccountId"}}.dkr.ecr.ap-southeast-2.amazonaws.com/{{values "service"}}:{{.BuildID}}'
+image: '{{values "prodAccountId"}}.dkr.ecr.{{values "region"}}.amazonaws.com/{{values "service"}}:{{.BuildID}}'
```
72636c, line 3: should this be

```suggestion
image: '{{values "prodAccountId"}}.dkr.ecr.{{.Region}}.amazonaws.com/{{values "service"}}:{{.BuildID}}'
```

or

```suggestion
image: '{{values "prodAccountId"}}.dkr.ecr.<%- region %>.amazonaws.com/{{values "service"}}:{{.BuildID}}'
```

template/koa-rest-api/.gantry/common.yml
```diff
@@ -1,11 +1,12 @@
...
 service: '<%- serviceName %>'
+region: '<%- prodAwsAccountId %>'
```
72636c, line 5:
```suggestion
```

72636c (approved):

template/koa-rest-api/.gantry/common.yml
bridgetbarnes: 2nd is clearer imo,  and more straightforward for people to change if its a cross region service

PR #498, opened 2021-08-29
Remove `@types/node` resolution override

72636c:
Jest 27.1 is compatible with newer versions of `@types/node`.


etaoins (approved):
I never thought I’d see this day arrive

PR #499, opened 2021-08-30
ci: remove yarn cache for worker dev deps image

samchungy:
The workers don't need to run `yarn install` after this step so we might as well remove the yarn cache associated with it.

In my testing -> this reduced the size of images by half

### DIFF
--- a/.changeset/wet-penguins-switch.md
+++ b/.changeset/wet-penguins-switch.md
@@ -1,0 +1,7 @@
+---
+'skuba': patch
+---
+
+**template:** Remove Yarn cache from worker Docker images
+
+This shrinks the cached Docker images that our worker templates generate.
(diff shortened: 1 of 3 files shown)


72636c (approved):
😍

Does the same apply to our non-worker templates, or is the cache required for the `yarn install --production` step seen elsewhere?

samchungy:
Yeah it's needed if yarn --offline is used, which is what the APIs seem to use when building the prod image

72636c:
Should we `&& yarn cache clean` off of e.g. this then?

https://github.com/seek-oss/skuba/blob/caeb86761fa254286c461b6092fa00f9af51d978/template/koa-rest-api/Dockerfile#L8

samchungy:
> Should we `&& yarn cache clean` off of e.g. this then?
> 
> https://github.com/seek-oss/skuba/blob/caeb86761fa254286c461b6092fa00f9af51d978/template/koa-rest-api/Dockerfile#L8

hmmm 🤔  I don't know Docker well enough, does the image layer from that step contribute to the final image? If so, maybe not?

72636c:
D'oh, that's me overlooking that we only copy `node_modules` over. We could include this in the Hello World template though, thoughts? https://github.com/seek-oss/skuba/blob/master/template/greeter/Dockerfile

samchungy:
> D'oh, that's me overlooking that we only copy `node_modules` over. We could include this in the Hello World template though, thoughts? https://github.com/seek-oss/skuba/blob/master/template/greeter/Dockerfile

Hmmm maybe, depends on that repos usage I guess

PR #503, opened 2021-09-01
Prime dev ECR cache in Lambda Buildkite pipelines

72636c:
I'm pretty sure this was unintentionally omitted as we already do this on a bunch of repos internally. Caught by @samchungy.


samchungy (approved):
🎉

PR #513, opened 2021-09-14
Remove experimental Babel support

72636c:


src/utils/manifest.ts
```diff
@@ -42,13 +42,3 @@ export const getEntryPointFromManifest = async () => {
     ? result.packageJson.skuba.entryPoint
     : DEFAULT_ENTRY_POINT;
 };
-
-export const isBabelFromManifest = async () => {
```
etaoins, line 46: Is it worth keeping this around so we can emit a warning? I’m not sure if anyone ended up using Babel support.

etaoins (approved):

src/utils/manifest.ts
72636c: I'm comfortable with a clean break here; I've fixed up the one instance of internal SEEK usage and this has always been marked as "experimental".

PR #519, opened 2021-09-15
Update lock file

AshwaryaSmridhi:
# Why is this change needed?
In this PR I am going to update the lock file in skuba because Jest has fixed the NodeJS.Glabal  namespace issue that started happening after @types/node  removed it.
Some of our repo is using the latest version of skuba but still seems to have the old jest version. I am not too sure about this but I am hoping update the lock file and making another skuba release might fix it 😅

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -776,7 +776,7 @@
     "@nodelib/fs.scandir" "2.1.5"
     fastq "^1.6.0"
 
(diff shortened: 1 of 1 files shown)


72636c:
Lockfiles aren't packaged and they have no effect on consumers, and observe that this refresh has not touched any Jest dependencies in any case.

Changes to our templates as a result of this fix were landed in #498.

PR #520, opened 2021-09-15
Start to break down README content

72636c:
We now have multiple pages and a Just the Docs-themed site.

---

I've deployed this from my local: https://seek-oss.github.io/skuba/


CONTRIBUTING.md
```diff
@@ -1,3 +1,7 @@
+---
+nav_order: 99
```
etaoins, line 2: 😮

etaoins (approved):

PR #523, opened 2021-09-15
Try GitHub Changesets changelog format

72636c:
This should be able to tag changes with their PR #. I'm not sure how this will work with the existing changelog entries but we'll have to wait and see.


etaoins (approved):
😮

tadhglewis (approved):

PR #525, opened 2021-09-16
Extend lint docs

72636c:
This incorporates some content from our internal Skutorial, as we discussed that we should place content closer to the actual tools at play.

### DIFF
--- a/docs/deep-dives/eslint.md
+++ b/docs/deep-dives/eslint.md
@@ -10,48 +10,94 @@ parent: Deep dives
 You may be familiar with it from SEEK's frontend development toolkit,
 [sku].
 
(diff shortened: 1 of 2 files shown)


docs/deep-dives/eslint.md
```diff
@@ -10,48 +10,94 @@ parent: Deep dives
...
+  extends: ['skuba'],
+  rules: {
+    // https://eslint.org/docs/rules/complexity
+    complexity: ['error', { max: 3 }],
+  },
+};
+```
+
+Let's check that our new rule has taken effect.
+Craft some absolute nonsense in a source file:
```
etaoins, line 40: 😄

etaoins (approved):

PR #526, opened 2021-09-16
Add changelog frontmatter during deploy

72636c:
Changesets doesn't like us mangling with the source CHANGELOG.md. See this brokenness: https://github.com/seek-oss/skuba/pull/496/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

---

I went this route because I'm not seeing an obvious way to configure how Changesets actually inserts into CHANGELOG.md, only formatting the individual entries beforehand: https://github.com/atlassian/changesets/blob/main/docs/modifying-changelog-format.md

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 set -e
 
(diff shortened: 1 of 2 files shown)


scripts/deploy.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
etaoins, line 1: Now we're getting really serious

etaoins (approved):

PR #539, opened 2021-09-21
Run ESLint and Prettier formatting in process

72636c:
This took an eternity because I went down a rabbit hole of parallelising Prettier and had little to show for it based on very rudimentary benchmarking of skuba itself; it seemed the overhead of spinning up multiple worker processes was significant. I may give this another shot with worker threads in future.

While this is undeniably more code that using the ESLint and Prettier CLIs out of the box, it presents a few opportunities:

- Customising logging output, particularly for `--debug`ging.
- Taking the above point further: supporting Buildkite annotations, particularly when `lint`ing.
- A marginally faster execution and smaller resource footprint.
- Likelier compatibility with alternative module resolution approaches like Yarn PnP, which can make it harder to reliably locate other `.bin` tools when `exec`ing.


integration/base/fixable/a/a/a.ts
```diff
@@ -0,0 +1,6 @@
+import path from 'path';
```
etaoins, line 1: For those without Lingersvision:

```suggestion
// Imports out-of-order
import path from 'path';
```

integration/base/ok/a/a/a.ts
```diff
@@ -0,0 +1,6 @@
+import fs from 'fs';
```
etaoins, line 1:
```suggestion
// Imports in order
import fs from 'fs';
```

src/cli/adapter/prettier.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const formatFile = async (
+  filepath: string,
+  logger: Logger,
+  mode: 'format' | 'lint',
+  result: Result,
+) => {
+  logger.debug(filepath);
+
+  const [config, data, fileInfo] = await Promise.all([
```
etaoins, line 27: I’m guessing this is using lower level helpers instead of `prettier.format` to give better user feedback? Is it worth leaving a comment to that effect?

src/cli/adapter/prettier.ts
```diff
@@ -0,0 +1,146 @@
...
+    touched: [],
+    unparsed: [],
+    untouched: [],
+  };
+
+  logger.debug('Processing files...');
+
+  const start = process.hrtime.bigint();
+
+  for (const filepath of filepaths) {
```
etaoins, line 116: What’s the performance difference of using `Promise.all` here? We could presumably overlap some I/O with the actual Prettier computation.

src/cli/adapter/prettier.ts
```diff
@@ -0,0 +1,146 @@
...
+import { crawlDirectory } from '../../utils/dir';
+import { Logger } from '../../utils/logging';
+import { getConsumerManifest } from '../../utils/manifest';
+
+interface Result {
+  count: number;
+  errored: Array<{ err?: unknown; filepath: string }>;
+  touched: string[];
+  unparsed: string[];
+  untouched: string[];
```
etaoins, line 16: Is this used anywhere?

etaoins (approved):

src/cli/adapter/prettier.ts
72636c: I've gone with a staged read->process->write approach so the debug logs remain in order.

src/cli/adapter/prettier.ts
72636c: `prettier.format` is fairly low level itself actually and is synchronous. I've tried to clarify this a little in the file.

etaoins (approved):

PR #542, opened 2021-09-22
Remove embarrassing directory reveal

72636c:
The hackers gon get me now


etaoins (approved):
I'll make sure your laptop is in scope for BugCrowd

PR #546, opened 2021-09-22
Make it easier to add changesets

72636c:
This makes it easier to use changeset-default formatting. We override the Prettier config to accept double quotes in the `.changeset` directory, and bold the commit title scope during changelog generation rather than requiring everyone to manually type out the asterisks.

### DIFF
--- a/.changeset/changelog.js
+++ b/.changeset/changelog.js
@@ -5,7 +5,7 @@
 const {
   getInfo,
   getInfoFromPullRequest,
-} = require('@changesets/get-github-info');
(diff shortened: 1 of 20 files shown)


.changeset/changelog.js
```diff
@@ -104,11 +104,14 @@ const changelogFunctions = {
       };
     })();
 
+    // Bold the scope.
```
etaoins, line 107: ❤️

etaoins (approved):

PR #556, opened 2021-09-24
build-package, lint: Add `--serial` flag

72636c:
Previously, we had a Buildkite-specific hack to propagate the BUILDKITE environment variable in order to run these commands serially. This is potentially surprising behaviour, particularly if the consumer is propagating the variable for some other purpose.

That point will become less hypothetical once skuba adds built-in support for Buildkite annotations. We replace the inference with a very explicit flag.

### DIFF
--- a/.changeset/itchy-cobras-bow.md
+++ b/.changeset/itchy-cobras-bow.md
@@ -1,0 +1,19 @@
+---
+"skuba": minor
+---
+
+build-package, lint: Add `--serial` flag
+
+This explicitly disables concurrent command execution.
+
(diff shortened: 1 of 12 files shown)


etaoins (approved):
Should we be updating the templates as well?

72636c:
@etaoins I go back and forth on that, but I'm not convinced that this should be the default. These commands can be significantly slower when run serially, and I feel right-sizing agents is worth the improvement in deployment time.

I'll try to specifically detect Gutenberg builds though since exhausting those agents causes much grief.

PR #557, opened 2021-09-24
Reset exit code between tests

72636c:
We lucked out by listing the exit code-setting test case last, but this will avoid future confusion.


etaoins (approved):
😌

PR #560, opened 2021-09-24
Add colour logo

72636c:
This is SVGed Cascadia Code, which seemed to be the only SIL font I had lying around that had properly smooth box-drawing characters.


etaoins (approved):
😍

PR #561, opened 2021-09-25
Generate per-template changelogs

72636c:
These are now available on each page under this section:

https://seek-oss.github.io/skuba/docs/templates/

---

This feature relies on a certain level of structure being applied to our templates and changelogs. To that end, we now validate in CI that template configuration is internally consistent and require all changelog entries to have a scope.

### DIFF
--- a/src/utils/template.ts
+++ b/src/utils/template.ts
@@ -7,6 +7,70 @@ import * as t from 'runtypes';
 
 import { ProjectType } from './manifest';
 
+export const TEMPLATE_NAMES = [
+  'express-rest-api',
+  'greeter',
+  'koa-rest-api',
+  'lambda-sqs-worker',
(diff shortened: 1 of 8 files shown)


src/utils/template.ts
```diff
@@ -7,6 +7,70 @@ import * as t from 'runtypes';
...
+
+export const TEMPLATE_NAMES_WITH_BYO = [...TEMPLATE_NAMES, 'github →'] as const;
+
+interface TemplateDocumentationConfig {
+  /**
+   * The semantic version in which the template was first added.
+   *
+   * This is used to filter out historical changelogs.
+   */
+  added: string;
```
etaoins, line 30: 💡

etaoins (approved):

PR #568, opened 2021-09-25
Escape backticks in Buildkite annotations

72636c:
This is paranoia in case some log happens to expose a sequence of 3+ backticks that could escape the terminal code block.


etaoins (approved):
Escaping text in annotation Markdown bitten me in the past as well. At least inside a fenced code block there's only a few ways to escape.

PR #569, opened 2021-09-25
Run changelog dry run without GITHUB_TOKEN

72636c:
IIRC keeping our validate workflow secret-free means that we can more easily run CI on forked contributions.

This also supports asterisk-based Markdown list items `* ` which seem to be generated by the GitHub-based formatter for some reason.


scripts/package.ts
```diff
@@ -164,6 +164,11 @@ const compileChangesByTemplate = (changelog: string) => {
               .replace(
                 /^([0-9a-f]{7,}): /,
                 (_, commit: string) => `${commitLink(commit)}: `,
+              )
+              // Auto-link a commit hash suffix to GitHub.
```
etaoins, line 168:
```suggestion
              // Auto-link a commit hash prefix to GitHub.
```
Or maybe "short commit hash"?

etaoins (approved):

scripts/package.ts
72636c: Ah 🤦, I was referring to it being suffixed on the end of the line, but I can see how this is confusing. Will fix.

PR #584, opened 2021-09-30
Limit Prettier to 25 parallel file I/O operations

72636c:
This seems to pop up on a large repo (300+ Prettified files) in Buildkite when our agents are under load.


etaoins (approved):
Ahh, this would do it 👍

Issue #398, opened 2021-04-03
Consider switching to ESM


72636c:
Some packages in the Node.js ecosystem are starting to go ESM-only, e.g. https://github.com/seek-oss/skuba/pull/397.

We can probably move most internal CLI code over to ESM without a breaking change; we already require Node.js 12+ and therefore ESM support via `"type": "module"` in `package.json`. The tricky part will be maintaining CommonJS compatibility in [src/api](https://github.com/seek-oss/skuba/tree/master/src/api) and [src/wrapper](https://github.com/seek-oss/skuba/tree/master/src/wrapper) as consumers call this code from their own module context and we don't want to force a switch at this time.

samchungy:
Last time I tried switching Jest -> Vitest it did not properly support CJS. So it seemed like I would have to both switch to ESM and Vitest at the same time which gives me lower confidence that the move worked.

samchungy:
https://github.com/seek-oss/skuba/blob/48c4208616e4dcd349670131fcad1fc9827e3ce7/src/skuba.ts#L48

This bit of code would need to be re-written as ESM doesn't support importing from directories/magic index.js files.

samchungy:
ESM requires file extensions within the imports.

1. We would either need to lint all of our existing code to add file extensions.

Or

2. Start bundling

samchungy:
Things that are blocking us:

1. Jest
  - Support is not there (https://github.com/jestjs/jest/issues/9430)
  - Moving to ESM would require everyone to change their mocks to [unstable mocks](https://jestjs.io/docs/ecmascript-modules#module-mocking-in-esm). eg. jest.unstable_mockModule. Unstable mocks do not support hoisting. File extensions in ESM are required but using them within Jest requires a [hack](https://github.com/kulshekhar/ts-jest/issues/1057#issuecomment-1482644543)

2. Skuba Dive
  - [module-alias does not work with ESM](https://github.com/ilearnio/module-alias/issues/99).

3. File Extensions
  - ESM requires file extensions to work, unless you introduce a bundler.

Current thinking:

1. Enable file extensions via an eslint rule while everyone is on CJS.

My rationale for this is, it should ultimately reduce the number of file changes once users switch to ESM. File extensions work within CJS but require a little Jest hack to get it to work.

2. Switch to Node16 Module Resolution

This is required for Step 3

[...]

PR #625, opened 2021-10-23
Add GitHub annotations for lint

samchungy:
## Features:
- Adds GitHub Annotations for lint to pull requests via the Check Runs API. (test will come in another PR as this PR is already relatively big)

## Changes:
- Expose raw Error and Warning results via Eslint adapter 
- Added new GitHub Annotations to Lint
- Added new GitHub API

#### Status View
image

#### Check Run Page View
image

#### Files Changed View
image


samchungy:
At the moment - I'm trying to work out how to output relative paths as Eslint currently emits absolute paths via the [Stylish formatter](https://eslint.org/docs/user-guide/formatters/#stylish).

Investigating if there is a way I can override the path programatically via the results in the eslint adapter or find a way for the Stylish formatter to produce relative paths.

https://github.com/eslint/eslint/issues/13376

image

src/cli/lint/annotate/github/eslint.ts
```diff
@@ -0,0 +1,29 @@
+import * as GitHub from '../../../../api/github';
+import { ESLintOutput } from '../../../../cli/adapter/eslint';
+
+export const createEslintAnnotations = (eslint: ESLintOutput) => {
+  const annotations: GitHub.Annotation[] = [];
+  [...(!eslint.ok ? eslint.errors : []), ...eslint.warnings].forEach(
```
72636c, line 6: This may read better as a `flatMap` rather than nested `forEach`es with `push`ing:

```typescript
export const createEslintAnnotations = (eslint) =>
  [...eslint.errors, ...eslint.warnings].flatMap(result =>
    result.messages.map(message => ({ /* formatting of said message */ })
  );
```

src/cli/lint/annotate/github/eslint.ts
```diff
@@ -0,0 +1,29 @@
...
+
+        annotations.push({
+          annotation_level: message.severity === 2 ? 'failure' : 'warning',
+          start_line: message.line || 1,
+          end_line: message.endLine || message.line || 1,
+          ...(column && { start_column: column }),
+          ...(endColumn && { end_column: endColumn }),
+          message: message.message,
+          path: result.filePath,
+          title: `Eslint${message.ruleId ? ` - ${message.ruleId}` : ''}`,
```
72636c, line 22: I'd have to see how this is rendered out, but I'm wondering if this would be better as

```suggestion
          title: `ESLint${message.ruleId ? ` (${message.ruleId})` : ''}`,
```

src/cli/lint/annotate/github/eslint.ts
```diff
@@ -0,0 +1,29 @@
...
+      result.messages.forEach((message) => {
+        // Annotations only support start_column and end_column on the same line.
+        const column = message.line === message.endLine && message.column;
+        const endColumn =
+          message.line === message.endLine && (message.endColumn || column);
+
+        annotations.push({
+          annotation_level: message.severity === 2 ? 'failure' : 'warning',
+          start_line: message.line || 1,
+          end_line: message.endLine || message.line || 1,
```
72636c, line 17: Are we trying to handle empty strings here, or can we use nullish coalescing `??`?

src/api/github/checkRun.ts
```diff
@@ -0,0 +1,119 @@
...
+        ]
+      : []),
+  ].join('\n\n');
+
+/**
+ * Creates a Check Run
+ * @param name - Name of the Check Run eg. lint
+ * @param summary - Summary of the report
+ * @param annotations - List of annotations
+ * @param conclusion - Conclusion of the run `failure` or `success`
```
72636c, line 87: I'd prefer these to be a parameter object. It's otherwise hard to eyeball what each parameter is in calling code, and easy to mix up `name` and `summary`.

docs/development-api/github.md
```diff
@@ -0,0 +1,38 @@
...
+parent: Development API
+---
+
+# GitHub
+
+---
+
+## createCheckRun
+
+Asynchronously creates a GitHub Check Run with annotations
```
72636c, line 11:
```suggestion
Asynchronously creates a GitHub Check Run with annotations.
```

Can we also link `Check Run` to some GitHub doco?

docs/development-api/github.md
```diff
@@ -0,0 +1,38 @@
...
+
+If the following environment variables are not present,
+the function will silently return without attempting to create a check run:
+
+- `BUILDKITE_REPO`
+- `BUILDKITE_COMMIT`
+- `BUILDKITE_BUILD_NUMBER`
+- `GITHUB_API_TOKEN`
+
```
72636c, line 20: It would also be good to document the maximum number of annotations and the behaviour of the function when `annotations` exceeds that.

72636c (commented):
Let me know if you need any help creating a beta release of this for internal testing 🙏

src/cli/adapter/eslint.ts
```diff
@@ -53,15 +60,29 @@ export const runESLint = async (
...
 
-  let errors = 0;
+  const errors: ESLintResult[] = [];
+  const warnings: ESLintResult[] = [];
 
   for (const result of results) {
-    errors += result.errorCount;
-
-    logger.debug(symbolForResult(result), path.relative(cwd, result.filePath));
+    const relativePath = path.relative(cwd, result.filePath);
```
72636c, line 67: 👍 This looks similar to https://github.com/seek-oss/skuba/blob/5dc3a2a4492f6095aaa850507d6eb3fd20361ec0/src/cli/adapter/eslint.ts#L61

Maybe we should have our adapter calculate its own `relativeFilePath` for reuse in the linked log message and here?

src/api/github/checkRun.ts
```diff
@@ -0,0 +1,122 @@
...
+    head_sha: process.env.BUILDKITE_COMMIT as string,
+    conclusion,
+  };
+
+  const result = await client.post<CreateCheckRunResponse>('/', {
+    data: createData,
+  });
+
+  // Add the other annotations to the result
+  await Promise.all(
```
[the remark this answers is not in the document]
72636c: I like your full summary, it's less mysterious 🙂

src/cli/lint/annotate/github/eslint.ts
samchungy: Yeah just handling `undefined`. I've switched it to the nullish coercing operator. I noticed while doing some manual testing that eslint parsing errors don't give us a start_line, so in order for the file to show up in the `Files Changed` tab we have to give it Line 1.

eg.

image

src/cli/lint/annotate/github/tsc.ts
```diff
@@ -0,0 +1,34 @@
+import { GitHub } from '../../../../';
+import { StreamInterceptor } from '../../../../cli/lint/external';
+
+const tscOutputRegex = new RegExp(
+  /^\x1B\[34mtsc      │\x1B\[39m (.*)\(([0-9]+),([0-9]+)\): error (.*)/,
```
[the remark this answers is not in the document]
samchungy: I've changed it to something I think will work:

`([^\s]*)\(([0-9]+),([0-9]+)\): error (.*)`. I've added some comments to describe what each capture group does in the code:

```
// Regex for typescript output lines. eg.
// "\x1B[34mtsc      │\x1B[39m src/index.ts(1,2): error TS6133: 'missing' is declared but its value is never read."
// Pulls out
// group 1: src/index.ts
// group 2: 1
// group 3: 2
// group 4: TS6133: 'missing' is declared but its value is never read.
```

src/cli/lint/annotate/github/eslint.ts
samchungy: Looks like this:

image

I'm indifferent. But I'll leave it as it is committed right now - with brackets

src/cli/adapter/eslint.ts
samchungy: A little bit confused at this one - but yeah I rewrote it to use the relativePath const

https://github.com/seek-oss/skuba/blob/9d97f1c9bb0b38ce37fbbe9be2c382d85f486583/src/cli/adapter/eslint.ts#L82

samchungy:
> Let me know if you need any help creating a beta release of this for internal testing 🙏

Maybe on Wednesday when I'm back 😄 I'll keep experimenting manually in the mean time to find little edge cases

src/cli/lint/annotate/github/eslint.ts
```diff
@@ -0,0 +1,26 @@
...
+
+        return {
+          annotation_level: message.severity === 2 ? 'failure' : 'warning',
+          start_line: message.line ?? 1,
+          end_line: message.endLine ?? message.line ?? 1,
+          ...(startColumn && { start_column: startColumn }),
+          ...(endColumn && { end_column: endColumn }),
+          message: message.message,
+          path: result.filePath,
+          title: `Eslint${message.ruleId ? ` (${message.ruleId})` : ''}`,
```
72636c, line 23:
```suggestion
          title: `ESLint${message.ruleId ? ` (${message.ruleId})` : ''}`,
```

src/cli/lint/annotate/github/tsc.ts
```diff
@@ -1,14 +1,20 @@
 import * as GitHub from '../../../../api/github';
 import { StreamInterceptor } from '../../../../cli/lint/external';
 
+const ansiPattern = [
```
samchungy, line 4: Copy pasted from `strip-ansi` and `ansi-regex` packages. Happy to import them instead if we want, however they are only esmodules and I couldn't quite get it working @72636c 

https://github.com/chalk/ansi-regex/blob/02fa893d619d3da85411acc8fd4e2eea0e95a9d9/index.js#L1

samchungy:
TSC Output needs a bit of work. Also very confused at the output - in one repo the error line/column is wrapped in brackets

```
tsc      │ src/workers/autoTopUpWorker/usecases/smokeTestInteractor.ts(20,7): error TS2769: No overload matches this call.
```

in another the output is 

```
tsc      │ src/functions/start-server/handler.ts:4:17 - error TS2305: Module '"src/logger"' has no exported member 'nope'.
```

Also some tsc errors are multiline 😕 eg.

```
tsc      │ src/workers/autoTopUpWorker/usecases/smokeTestInteractor.ts(20,7): error TS2769: No overload matches this call.
tsc      │   Overload 1 of 2, '(obj: LogContext, msg?: string | undefined, ...args: any[]): void', gave the following error.
tsc      │     Argument of type 'unknown' is not assignable to parameter of type 'LogContext'.
tsc      │   Overload 2 of 2, '(msg?: string | undefined, ...args: any[]): void', gave the following error.
tsc      │     Argument of type 'unknown' is not assignable to parameter of type 'string | undefined'.
tsc      │       Type 'unknown' is not assignable to type 'string'.
```

Need to do some thinking about how to do this.

samchungy:
Okay it looks like when the `"pretty": true` option is provided in `compilerOptions` the output changes to colons separating to the colon syntax

samchungy:
Going to take inspiration from https://github.com/actions/setup-node/pull/125/files to see what they do

samchungy:
Okay - I've managed to work together a regex which can handle both the pretty output and the regular output and also cover multi lines. I'll add some unit test coverage tomorrow 😆

docs/development-api/github.md
```diff
@@ -0,0 +1,41 @@
+---
+parent: Development API
+---
+
+# GitHub
+
+---
+
+## createCheckRun
```
72636c, line 9: This has me a bit conflicted; while this is the simplest API to support typical SEEK usage, I wonder if it's too Buildkite-specific for the name.

What do you think about:

1. Having a more wordy name like `createCheckRunFromBuildkite` to qualify that this is tailored towards Buildkite usage

2. Accepting an optional `auth` or `githubApiToken` parameter that falls back to the environment variable, so it still works out of the box for us but isn't intrinsically tied to SEEK's custom `GITHUB_API_TOKEN`?

   (I haven't actually convinced myself that this is a good idea, and maybe we can leave it until someone asks given it would be a non-breaking addition.)

docs/development-api/github.md
```diff
@@ -0,0 +1,41 @@
...
+Asynchronously creates a GitHub [Check Run] with annotations.
+
+If the number of `annotations` provided exceeds 50, the number sent to GitHub will be capped to 50 and an explanation will be appended to the `summary`.
+
+If the following environment variables are not present,
+the function will silently return without attempting to create a check run:
+
+- `BUILDKITE_REPO`
+- `BUILDKITE_COMMIT`
+- `BUILDKITE_BUILD_NUMBER`
```
72636c, line 20:
```suggestion
- `BUILDKITE_BUILD_NUMBER`
- `BUILDKITE_COMMIT`
- `BUILDKITE_REPO`
```

src/cli/lint/annotate/github/tsc.ts
72636c: I think we can use an older major version here. Only the latest should be ESM-only.

Migrating skuba itself to ESM is something we can tackle later and is tracked in #398.

72636c (commented):
Mostly doco nitpicks 😅, thanks for bearing with me

docs/development-api/github.md
samchungy: 1. Happy to use `createCheckRunFromBuildkite` given we are relying on `BUILDKITE_` environment variables.

2. 🤔 Hmmm I'm going to lean towards leaving it til someone requests it.

src/cli/lint/annotate/github/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  summary: string,
+) => {
+  const annotations: GitHub.Annotation[] = [
+    ...createEslintAnnotations(eslint),
+    ...createPrettierAnnotations(prettier),
+    ...createTscAnnotations(tscOk, tscOutputStream),
+  ];
+
+  const isOk = eslint.ok && prettier.ok && tscOk;
+  const conclusion = isOk ? 'success' : 'failure';
```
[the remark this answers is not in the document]
samchungy: If there is a fail in a previous run we need to overwrite it with a ✅  or else there will be a status check with a failure on the PR 😛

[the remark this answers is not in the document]
72636c: Interesting, the check run persists across commits to the PR?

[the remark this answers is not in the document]
samchungy: Ooo actually, no. Just tested it. I've been overwriting the same commit 😆 

I think it would be good to keep to know that it's working though?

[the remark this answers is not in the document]
72636c: Yep fair 👍

src/api/github/checkRun.ts
```diff
@@ -0,0 +1,119 @@
...
+    return;
+  }
+
+  const client = new Octokit({
+    auth: process.env.GITHUB_API_TOKEN,
+  });
+  const { owner, repo } = getOwnerRepo();
+  const title = createTitle(conclusion, annotations.length);
+  const enrichedSummary = createEnrichedSummary(summary, annotations.length);
+  const checkRunName = `skuba/${name}`;
```
72636c, line 106: IMO this magic doesn't really belong here and we should pass through the `name` parameter verbatim. Users will be able to use this function to write their own arbitrary check runs, and those wouldn't really make sense under a `skuba/` namespace.

src/api/github/checkRun.ts
```diff
@@ -0,0 +1,119 @@
...
+  ].join('\n\n');
+
+interface CreateCheckRunParameters {
+  name: string;
+  summary: string;
+  annotations: Annotation[];
+  conclusion: 'failure' | 'success';
+}
+
+export const createCheckRunFromBuildkite = async ({
```
72636c, line 90:
```suggestion
/**
 * Asynchronously creates a GitHub [check run] with annotations.
 *
 * This writes the first 50 `annotations` in full to GitHub.
 *
 * If the following environment variables are not present,
 * the function will silently return without attempting to annotate:
 *
 * - `BUILDKITE_BUILD_NUMBER`
 * - `BUILDKITE_COMMIT`
 * - `BUILDKITE_REPO`
 * - `GITHUB_API_TOKEN`
 */
export const createCheckRunFromBuildkite = async ({
```

src/cli/lint/annotate/github/eslint.test.ts
```diff
@@ -0,0 +1,203 @@
...
+      start_line: 4,
+      title: 'ESLint (@typescript-eslint/no-unused-vars)',
+    },
+  ];
+  const annotations = createEslintAnnotations(eslintOutput);
+
+  expect(annotations).toStrictEqual(expectedAnnotations);
+});
+
+it('should not annotate column and end column only if start and end line are the same', () => {
```
72636c, line 164:
```suggestion
it('should not specify columns when an annotation spans multiple lines', () => {
```

src/api/github/checkRun.ts
```diff
@@ -0,0 +1,120 @@
...
+      process.env.BUILDKITE_COMMIT &&
+      process.env.BUILDKITE_REPO &&
+      process.env.GITHUB_API_TOKEN,
+  );
+
+// Pulls out the GitHub Owner + Repo String from repo urls eg.
+// git@github.com:seek-oss/skuba.git
+// https://github.com/seek-oss/skuba.git
+// Pulls out `seek-oss` as owner and `skuba` as repo
+const ownerRepoRegex = new RegExp(/github.com(?::|\/)(.*)\/(.*).git/);
```
72636c, line 24:
```suggestion
/**
 * Matches the owner and repository names in a GitHub repository URL.
 *
 * For example, given the following input strings:
 *
 * ```console
 * git@github.com:seek-oss/skuba.git
 * https://github.com/seek-oss/skuba.git
 * ```
 *
 * This pattern will produce the following matches:
 *
 * 1. seek-oss
 * 2. skuba
*/
const ownerRepoRegex = new RegExp(/github.com(?::|\/)(.+)\/(.+).git$/);
```

src/api/github/checkRun.ts
```diff
@@ -0,0 +1,120 @@
...
+    owner,
+    repo,
+  };
+};
+
+/**
+ * Create a uniform title format for our check runs
+ * @param conclusion - `failure` or `success`
+ * @param annotationsLength - Number of annotations added
+ * @returns Title eg. Build #12 failed (24 annotations added)
```
72636c, line 47:
```suggestion
 * Create a uniform title format for our check runs, e.g.
 *
 * ```text
 * Build #12 failed (24 annotations added)
 * ```
```

src/api/github/checkRun.ts
```diff
@@ -0,0 +1,120 @@
...
+ */
+const createTitle = (
+  conclusion: 'failure' | 'success',
+  annotationsLength: number,
+): string => {
+  const build = `Build #${process.env.BUILDKITE_BUILD_NUMBER as string}`;
+  const numAnnotations =
+    annotationsLength > GITHUB_MAX_ANNOTATIONS
+      ? GITHUB_MAX_ANNOTATIONS
+      : annotationsLength;
```
72636c, line 57: How about `inputAnnotations` and `addedAnnotations` as names?

src/api/github/checkRun.ts
```diff
@@ -0,0 +1,120 @@
...
+  const plural = numAnnotations === 1 ? '' : 's';
+  return `${build} ${status} (${numAnnotations} annotation${plural} added)`;
+};
+
+/**
+ * Adds more context to the summary provided
+ * @param summary - report summary
+ * @param annotationsLength - Number of annotations added
+ * @returns summary with extra metadata
+ */
```
72636c, line 69:
```suggestion
/**
 * Enriches the summary with more context about the check run.
 */
```

src/api/github/checkRun.ts
```diff
@@ -0,0 +1,120 @@
...
+const createEnrichedSummary = (
+  summary: string,
+  annotationsLength: number,
+): string =>
+  [
+    summary,
+    ...(annotationsLength > GITHUB_MAX_ANNOTATIONS
+      ? [
+          `There were ${annotationsLength} annotations created. However, the number of annotations displayed has been capped to ${GITHUB_MAX_ANNOTATIONS}`,
```
72636c, line 78:
```suggestion
          `${inputAnnotations} annotations were provided, but only the first ${GITHUB_MAX_ANNOTATIONS} are visible in GitHub.`,
```

72636c (commented):
This is outstanding 🌟

72636c (approved):

src/cli/lint/annotate/github/tsc.ts
```diff
@@ -0,0 +1,70 @@
...
+export const createTscAnnotations = (
+  tscOk: boolean,
+  tscOutputStream: StreamInterceptor,
+): GitHub.Annotation[] => {
+  const annotations: GitHub.Annotation[] = [];
+  if (!tscOk) {
+    const rawOutput = stripAnsi(tscOutputStream.output()).replace(
+      tscPrefixRegex,
+      '',
+    );
```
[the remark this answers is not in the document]
samchungy: Okay I believe we could strip the tsc prefix there but I think and correct me I am wrong because I have not used it but BuildKite Annotations I think can make use of the Ansi colouring?

src/cli/lint/annotate/github/tsc.ts
```diff
@@ -0,0 +1,70 @@
...
+  'g',
+);
+
+const annotationLevelMap: Record<
+  TscLevel,
+  GitHub.Annotation['annotation_level']
+> = {
+  error: 'failure',
+  warning: 'warning',
+  info: 'notice',
```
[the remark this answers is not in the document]
samchungy: Nope hehe but the Regex which [vscode uses](https://github.com/microsoft/vscode/blob/0fb0d01367872e576c95a755fe53fb8a73b3cc6a/extensions/typescript-language-features/package.json#L1290) - uses it. I just extended it to capture multiple lines after the output. So I would assume there are info and warnings in a similar shape

src/cli/lint/annotate/github/tsc.ts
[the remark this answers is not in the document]
72636c: Ah great point 👍

src/cli/lint/annotate/github/tsc.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+
+export const createTscAnnotations = (
+  tscOk: boolean,
+  tscOutputStream: StreamInterceptor,
+): GitHub.Annotation[] => {
+  const annotations: GitHub.Annotation[] = [];
```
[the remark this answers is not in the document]
samchungy: image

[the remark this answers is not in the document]
72636c: Ah, that's annoying 😞

```typescript
const matches = stripAnsi(output()).matchAll(regex);

return Array.from(matches).flatMap();
```

src/cli/lint/annotate/github/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  const reportSummary = isOk ? 'Lint passed' : summary;
```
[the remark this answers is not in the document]
samchungy: Hmmm not quite sure I follow you here - could you provide an example?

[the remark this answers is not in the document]
72636c: Ah, I meant that this code is a bit weird overall. We take in a `summary` parameter that is discarded on success (`Lint passed`), and we have all the information necessary from the other parameters to summarise the failure ourselves.

I reckon we can drop the parameter and go with something like this:

```suggestion
  const summary = isOk
    ? '`skuba lint` passed.'
    : '`skuba lint` found issues that require triage.';
```

And in future if we really wanted to, we could expand the failure summary to include detail about the number of ESLint errors and warnings, un-Prettified files, etc.

72636c (approved):
:shipit: thank you!

PR #640, opened 2021-11-03
Switch `lambda-sqs-worker` template to ARM64

etaoins:
These are a bit cheaper and a bit faster than the x86 Lambdas:
<https://aws.amazon.com/blogs/aws/aws-lambda-functions-powered-by-aws-graviton2-processor-run-your-functions-on-arm-and-get-up-to-34-better-price-performance/>

The underlying Lambda architecture should be invisible to typical TypeScript Lambdas. Indirect Apply has switched over all of their Lambdas, only encountering difficulty with DataDog/serverless-plugin-datadog#184.

### DIFF
--- a/.changeset/clever-masks-develop.md
+++ b/.changeset/clever-masks-develop.md
@@ -1,0 +1,10 @@
+---
+'skuba': minor
+---
+
+**template/lambda-sqs-worker:** Switch to Graviton2
+
+These are a bit cheaper and a bit faster than the x86 Lambdas:
+<https://aws.amazon.com/blogs/aws/aws-lambda-functions-powered-by-aws-graviton2-processor-run-your-functions-on-arm-and-get-up-to-34-better-price-performance/>
+
+The underlying Lambda architecture should be invisible to typical TypeScript Lambdas.
--- a/.changeset/clever-masks-develop.md
+++ b/.changeset/clever-masks-develop.md
@@ -1,5 +1,5 @@
 ---
-'skuba': minor
+"skuba": minor
 ---
 
 **template/lambda-sqs-worker:** Switch to ARM64 architecture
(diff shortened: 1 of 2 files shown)


.changeset/clever-masks-develop.md
```diff
@@ -0,0 +1,10 @@
+---
+"skuba": minor
```
72636c, line 2:
```suggestion
"skuba": patch
```

.changeset/clever-masks-develop.md
```diff
@@ -0,0 +1,10 @@
...
+---
+
+**template/lambda-sqs-worker:** Switch to ARM64 architecture
+
+These are a bit cheaper and a bit faster than the x86 Lambdas:
```
72636c, line 7:
```suggestion
These are a bit cheaper and a bit faster than x86 Lambdas:
```

72636c (approved):

samchungy:
I think it might be worth leaving a comment somewhere that you should be using arm64 build agents since buildx isn't available on our regular build agents yet

etaoins:
This doesn't require ARM64 build agents. Serverless uploads a source ZIP, there's not binaries involved.

samchungy:
However - when we run eg. `yarn`/`yarn install` when building dependencies - won't native deps require ARM64 binaries? As I understand the `arm64` flag is just for telling AWS Lambda that the bundle contains arm64 deps?

etaoins:
You can't use native dependencies on the `nodejs14.x` runtime. It's a special Node environment that's maintained by Amazon and won't execute arbitrary native code.

To include native node deps on Lambda, you'd need to use something like SAM to build a Docker image:
<https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-package.html#gettingstarted-package-awsother>

If you're going to that length, you need to change the `runtime` anyway so updating the `architecture` is trivial.

samchungy:
Gotcha - thanks for the explanation! 😄

PR #641, opened 2021-11-06
Decouple GitHub.createCheckRun from Buildkite

72636c:
1. Source GIt metadata from Git repo.

   This reduces the number of Buildkite environment variables required to use `GitHub.createCheckRunFromBuildkite`, while adding the basic `BUILDKITE` variable for symmetry with `Buildkite.annotate`.

2. Move Buildkite specifics out of the GitHub development API and into the lint adapter.

   This provides a cleaner generic interface for our API. We also generalise the adapter code so we can support other CI platforms like GitHub Actions.

### DIFF
--- a/.changeset/cuddly-forks-give.md
+++ b/.changeset/cuddly-forks-give.md
@@ -2,4 +2,4 @@
 "skuba": minor
 ---
 
-GitHub.createCheckRunFromBuildkite: Add development API for writing annotations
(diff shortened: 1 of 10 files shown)


samchungy (approved):
Ahhhh smarty! 🤓 I like it!

etaoins (approved):
🎉

PR #648, opened 2021-11-07
GitHub Test Annotations

### LINKED ISSUE #638: GitHub Annotations for test
Add [GitHub Annotations](https://github.com/seek-oss/skuba/issues/619) for `skuba test`.

### Users may have multiple tests (unit/integration) they run with Jest, How do we decide what `name` to use for the Check Run - `skuba/${name}` eg. `skuba/test`

- Use Jest Projects display name
- Environment variable
- CLI flag?
- Pass an options argument to a [reporter](https://jestjs.io/docs/configuration#reporters-arraymodulename--modulename-options)
```
{
  "reporters": [
    "default",
    ["<rootDir>/my-custom-reporter.js", {"banana": "yes", "pineapple": "no"}]
  ]
}
```

### How do we implement annotations?

- Regex on Jest output
- Implement a Jest Reporter
  - Separate repo or js file reporter

samchungy:
Resolves #638

## Description
Adds a custom Jest Reporter which implements the new GitHub annotations API. This adds annotations to the `skuba test` command.

This will create a separate check run per jest.config displayName


samchungy:
There's a slight issue with this approach and will need some good documentation around it.

At the moment if a project uses only a single `jest.config` file but uses different CLI commands for different tests. eg. `yarn test --testPatterns *.int.test.ts` and `yarn test --testPatterns *.unit.test.ts`. There will only be one check run with the name `skuba test` created with the latter call overwriting the results of the first. I cannot think of a simple workaround for this unless we call to get all the check runs and create one with a different name eg. `skuba test 2`.

I would like to encourage users to create separate `jest.config` files for different test types eg. to achieve the same as above - create separate jest.config files.

Eg. 
jest.config.int.ts
```
export default {
  ...Jest.mergePreset({
    displayName: 'integration',
    testPatterns: ['*.int.test.ts']
  }),
};
```

jest.config.ts
```
export default {
  ...Jest.mergePreset({
    displayName: 'unit',
    testPatterns: ['*.unit.test.ts']
  }),
};
```

This will create two separate check runs `skuba/test (integration)` and `skuba/test (unit)`.

Alternative, you could choose to not label one with a displayName. 

Eg. 
```
export default {
  ...Jest.mergePreset({
    displayName: 'integration',
    testPatterns: ['*.int.test.ts']
  }),
};
```

jest.config.ts
```
export default {
  ...Jest.mergePreset({
    testPatterns: ['*.unit.test.ts']
  }),
};
```
This will alsos create two separate check runs `skuba/test (integration)` and `skuba/test`.

src/cli/test/reporters/github/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    const annotationResults = Object.entries(sortedTestResults).map<{
+      displayName: string;
+      annotations: GitHub.Annotation[];
+    }>(([displayName, testResults]) => ({
+      displayName,
+      annotations: createAnnotations(testResults),
+    }));
+
+    // Create a check run per display name. Run in series.
+    await pMap(
```
72636c, line 42: If we're running these serially, I think we'd be better off with a plain old `for...of` loop.

jest-preset.js
```diff
@@ -44,4 +44,8 @@ module.exports = {
     '/node_modules.*/',
     '<rootDir>/(coverage|dist|lib|tmp).*/',
   ],
+  reporters: [
+    'default',
+    '<rootDir>/node_modules/skuba/lib/cli/test/reporters/github',
```
72636c, line 49: How about we `require.resolve` this path? That should allow for internal use, though you'd still have to build `skuba` itself beforehand. It should also serve as an integration test of sorts as I assume our CI will break if we fail to resolve this reporter.

src/cli/test/reporters/github/index.ts
```diff
@@ -0,0 +1,66 @@
+import { Context, Reporter } from '@jest/reporters';
+import { AggregatedResult, TestResult } from '@jest/test-result';
```
72636c, line 2: This may make it clearer that we're not taking a runtime dependency on these subpackages:

```suggestion
import type { Context, Reporter } from '@jest/reporters';
import type { AggregatedResult, TestResult } from '@jest/test-result';
```

src/cli/test/reporters/github/annotations.ts
```diff
@@ -0,0 +1,51 @@
+import path from 'path';
+
+import { TestResult } from '@jest/test-result';
+
+const jestRegex = /\((.+?):(\d+):(\d+)\)/;
+
+import * as GitHub from '../../../../api/github';
+
+const createAnnotations = (testResults: TestResult[]) => {
```
72636c, line 9:
```suggestion
export const createAnnotations = (testResults: TestResult[]) => {
```

src/cli/test/reporters/github/annotations.ts
```diff
@@ -0,0 +1,51 @@
...
+  const cwd = process.cwd();
+
+  const annotations: GitHub.Annotation[] = [];
+  testResults.forEach((testResult) => {
```
72636c, line 13: I wonder if we can work some `flatMap`ping in here too?

72636c (commented):
Exciting 🙌

72636c:
I think this `displayName`-based approach makes sense. We can either encourage users to have multiple config files with distinct `displayName`s, or a single config file with multiple `projects` and associated `displayName`s (if the latter, we should think about how we can better support that style in `Jest.mergePreset`).

We could hash `process.cwd()` and `process.argv` if we wanted to completely avoid clobbering check runs, but I wouldn't bother with that unless we see this `displayName` approach fall over in practice. Keep in mind this won't enable itself on existing repos until peeps update them to propagate the necessary prerequisites, and they can fix up configs while they are there anyway.

src/cli/test/reporters/github/index.ts
```diff
@@ -0,0 +1,66 @@
...
+          ? '`skuba test` passed.'
+          : '`skuba test` found issues that require triage.';
+        const build = buildNameFromEnvironment();
+
+        await GitHub.createCheckRun({
+          name,
+          annotations,
+          conclusion,
+          summary,
+          title: `${build} ${isOk ? 'passed' : 'failed'}`,
```
72636c, line 60: Not for this PR, but I'd be interested to see whether we can extend this with coverage in future. It would be neat if we could compare against the default branch a la [Codecov](https://docs.codecov.com/docs/pull-request-comments); could we try to retrieve details of the equivalent check run on the head commit for comparison?

samchungy: Might be able to grab it from the test result https://github.com/facebook/jest/blob/8f2cdad7694f4c217ac779d3f4e3a150b5a3d74d/packages/jest-test-result/src/types.ts#L95

src/cli/test/reporters/github/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  buildNameFromEnvironment,
+  enabledFromEnvironment,
+} from '../../../../api/github/environment';
+
+import { createAnnotations } from './annotations';
+
+const DEFAULT_NAME = 'noDisplayName';
+
+export default class GitHubReporter implements Pick<Reporter, 'onRunComplete'> {
+  async onRunComplete(
```
72636c, line 15: Does Jest fail the test run if a reporter throws? If so, we'll need to roll something like #649 ourselves.

samchungy: Good question - I should test that

src/cli/test/reporters/github/index.ts
```diff
@@ -0,0 +1,67 @@
...
+          name,
+          annotations,
+          conclusion,
+          summary,
+          title: `${build} ${isOk ? 'passed' : 'failed'}`,
+        });
+      }
+    } catch (err) {
+      log.warn('Failed to annotate results.');
+      log.warn(err);
```
72636c, line 64: 👏

docs/cli/test.md
```diff
@@ -22,4 +22,21 @@ Arguments are passed through to the Jest CLI:
 skuba test --coverage path/to/file.test.ts
 ```
 
+### Annotations
+
+`skuba lint` can automatically emit annotations in CI.
```
samchungy, line 27:
```suggestion
`skuba test` can automatically emit annotations in CI.
```

src/cli/test/reporters/github/index.ts
```diff
@@ -0,0 +1,51 @@
...
+      return;
+    }
+
+    try {
+      const entries = generateAnnotationEntries(testResults);
+
+      const build = buildNameFromEnvironment();
+
+      // Create a check run per display name.
+      // Run in series to reduce the likelihood of exceeding GitHub rate limits.
```
72636c, line 28: Just double checking that this was the intent of running in series?

72636c (approved):
🙇

src/cli/test/reporters/github/index.ts
samchungy: Correcto, you wake up early as 👀

PR #654, opened 2021-11-08
Fix docker compose file environment key in templates

samchungy:
Builds with the current docker-compose.yml files will throw this:

```
ERROR: The Compose file './docker-compose.yml' is invalid because:
--
  | Unsupported config option for services.build-schema: 'env'

```


72636c (approved):
🤦 thank you

PR #664, opened 2021-11-11
Fix file descriptor warnings

### LINKED ISSUE #659: Prettier Warning/Errors
https://buildkite.com/seek/hirer-contracts-api/builds/1972#

Recently bumped up an API to Skuba 3.15.2. We got this error when running `yarn lint`.
image

Re-running it again worked however we get left with some warnings:

```
(node:30) Warning: File descriptor 32 closed but not opened in unmanaged mode
--
  | (Use `node --trace-warnings ...` to show where the warning was created)
  | Prettier │ Processed 85 files in 10.27s.

```

Ran --trace-warnings on local and got the following:

```
➜ node --trace-warnings node_modules/.bin/skuba lint                                                                                                                                           
(node:33523) Warning: File descriptor 43 closed but not opened in unmanaged mode
    at Object.closeSync (node:fs:530:11)
    at Object.closeSync (/Users/schung/repos/hirer-contracts-api/node_modules/graceful-fs/graceful-fs.js:74:20)
    at getInterpreter (/Users/schung/repos/hirer-contracts-api/node_modules/prettier/index.js:12674:30)
    at Object.inferParser (/Users/schung/repos/hirer-contracts-api/node_modules/prettier/index.js:12691:25)
    at getFileParser (/Users/schung/repos/hirer-contracts-api/node_modules/prettier/index.js:25444:36)
    at _getFileInfo (/Users/schung/repos/hirer-contracts-api/node_modules/prettier/index.js:25484:29)
    at getFileInfo (/Users/schung/repos/hirer-contracts-api/node_modules/prettier/index.js:25407:10)
    at async Promise.all (index 2)
```

72636c:
Resolves #659.

### DIFF
--- a/src/cli/adapter/prettier.ts
+++ b/src/cli/adapter/prettier.ts
@@ -37,7 +37,7 @@ let languages: SupportLanguage[] | undefined;
  * - seek-oss/skuba#659
  *
  */
-const inferParser = (filepath: string) => {
+export const inferParser = (filepath: string) => {
   const filename = path.basename(filepath).toLowerCase();
 
   languages ??= getSupportInfo().languages.filter((language) => language.since);
--- a/src/cli/adapter/prettier.ts
+++ b/src/cli/adapter/prettier.ts
@@ -35,9 +35,8 @@ let languages: SupportLanguage[] | undefined;
  *
  * - https://github.com/prettier/prettier/blob/2.4.1/src/main/options.js#L167
  * - seek-oss/skuba#659
- *
  */
-export const inferParser = (filepath: string) => {
+export const inferParser = (filepath: string): string | undefined => {
   const filename = path.basename(filepath).toLowerCase();
 
   languages ??= getSupportInfo().languages.filter((language) => language.since);
--- a/src/cli/adapter/prettier.ts
+++ b/src/cli/adapter/prettier.ts
@@ -1,11 +1,55 @@
 import path from 'path';
 
 import fs from 'fs-extra';
-import { Options, check, format, getFileInfo, resolveConfig } from 'prettier';
+import {
+  Options,
+  SupportLanguage,
+  check,
+  format,
+  getSupportInfo,
+  resolveConfig,
+} from 'prettier';
 
 import { crawlDirectory } from '../../utils/dir';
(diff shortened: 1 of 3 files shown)


src/cli/adapter/prettier.ts
```diff
@@ -1,12 +1,56 @@
...
+  format,
+  getSupportInfo,
+  resolveConfig,
+} from 'prettier';
 
 import { crawlDirectory } from '../../utils/dir';
 import { Logger, pluralise } from '../../utils/logging';
 import { getConsumerManifest } from '../../utils/manifest';
 
+let languages: SupportLanguage[] | undefined;
```
samchungy, line 17: question: is there a reason we are using a global variable?

src/cli/adapter/prettier.ts
```diff
@@ -1,12 +1,56 @@
...
+ *     ...
+ * ```
+ *
+ * References:
+ *
+ * - https://github.com/prettier/prettier/blob/2.4.1/src/main/options.js#L167
+ * - seek-oss/skuba#659
+ *
+ */
+export const inferParser = (filepath: string) => {
```
samchungy, line 40: suggestion: add return type?

samchungy (commented):
woah so quick 🏃 💨

etaoins (approved):
Thanks for hunting this down! I was hoping there would be a nicer fix but I’m not surprised this is the best we can do.

Are there any cases where the missing shebang detection would hurt us on a typical SEEK codebase?

72636c:
@etaoins I can't imagine there would be much impact here. Shebang detection is a fallback that is only invoked where file root and extension matching have turned up empty, and the filename does not have any extension. Prettier doesn't handle typical shell scripting languages; we could possibly lose out on `zx` but if that picks up steam at SEEK I'm sure we can special case it in our code.

src/cli/adapter/prettier.ts
72636c: This memoises the `languages` array rather than recalculating it on every file that passes through `inferParser`. We don't vary the plugins between files so the array is always the same.

PR #668, opened 2021-11-16
Enforce consistent type imports

72636c:
Fun times. This is auto-fixable and is internal to skuba for now.

https://github.com/typescript-eslint/typescript-eslint/blob/v5.4.0/packages/eslint-plugin/docs/rules/consistent-type-imports.md


etaoins (approved):
This is neat

PR #670, opened 2021-11-16
Separate git utils from checkRun

samchungy:
We need to use the owner/repo function elsewhere. Separated the common functions we may use elsewhere into their own file in the utils folder.


src/api/github/checkRun.test.ts
```diff
@@ -51,12 +50,13 @@ describe('createCheckRun', () => {
 
   beforeEach(() => {
     mocked(Octokit).mockReturnValue(mockClient as unknown as Octokit);
-    mocked(git.listRemotes).mockResolvedValue([
```
72636c, line 54: Can we keep these as-is? I know opinions differ here, but I like to think we gain a bit more integration assurance that way rather than going full unit.

src/utils/git.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const result = await getOwnerRepo(dir);
+    expect(result).toStrictEqual({
+      owner: 'seek-oss',
+      repo: 'skuba',
+    });
+  });
+});
+
+describe('getHeadSha', () => {
+  it('should extract a GitHub owner and repo from Git remotes', async () => {
```
72636c, line 35:
```suggestion
  it('should extract a commit hash from the Git log', async () => {
```

.changeset/pink-gifts-count.md
```diff
@@ -0,0 +1,5 @@
+---
```
72636c, line 1: We can do without a changeset here since we aren't changing the public interface; `/src/utils` should be all private.

72636c (approved):

src/api/github/checkRun.test.ts
samchungy: sure

src/utils/git.test.ts
samchungy: oops 😬

samchungy: Caught copy pasting 😆

Issue #652, opened 2021-11-07
Annotate test coverage


72636c:
This is a logical extension of #638. From https://github.com/seek-oss/skuba/pull/648#discussion_r744232683:

> It would be neat if we could compare against the default branch a la [Codecov](https://docs.codecov.com/docs/pull-request-comments); could we try to retrieve details of the equivalent check run on the head commit for comparison?

PR #675, opened 2021-11-16
Report on test coverage

72636c:
This is just the first step of dumping a coverage table out to a Buildkite annotation and GitHub check run. We can later build towards:

- Providing Buildkite annotations of failing tests
- Providing test coverage as a GitHub PR comment
- Comparing test coverage against the default branch per #652


src/cli/test/reporters/github/index.ts
```diff
@@ -35,11 +40,29 @@ export default class GitHubReporter implements Pick<Reporter, 'onRunComplete'> {
           ? '`skuba test` passed.'
           : '`skuba test` found issues that require triage.';
 
+        if (coverage) {
+          const buildkiteOutput: string = [
+            '`skuba test` coverage:',
+            Buildkite.md.terminal(coverage),
+          ].join('\n\n');
+
+          await Buildkite.annotate(buildkiteOutput, {
```
72636c, line 49: @samchungy how do you feel about having this reporter support both Buildkite and GitHub 😬

If we want coverage integrations across both, it may be wasteful to define two reporters and to run `renderCoverageText` twice.

samchungy: I don't love it but `renderCoverageText` is kinda costly I guess.

Do reporters run in series? I wonder if we could store it in memory? 🤔

If we do this -> might want to rename the reporter?

samchungy:
Maybe something we can add to https://github.com/seek-oss/skuba/issues/2345

PR #678, opened 2021-11-17
Register `tsconfig-paths` for development

72636c:
Note that this does not modify our typical `skuba-dive/register` runtime hook. The drift between the two may be a bit puzzling in the time being, but it's somewhat rare for production applications to include `tsconfig.json` in their final bundle or container. Alignment should eventually come as we investigate shifting over to `esbuild`, which supports compile-time `tsconfig-paths` resolution.

### DIFF
--- a/.changeset/funny-snails-share.md
+++ b/.changeset/funny-snails-share.md
@@ -1,0 +1,19 @@
+---
+"skuba": minor
+---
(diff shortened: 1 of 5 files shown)


etaoins (approved):
❤️

PR #681, opened 2021-11-17
Start on experimental esbuild support

### LINKED ISSUE #588: Investigate esbuild for `skuba build`
This presents a few opportunities while still preserving partial compatibility with our existing `tsconfig.json`-based builds with `tsc`:

- Bundle code to improve cold start time, particularly in AWS Lambda
- Extend the build process with plugins, including build-time module alias resolution
- In-process execution, as it supports a convenient Node.js API while `tsc` is CLI-only
- Faster builds, as it skips type checking which we already handle on lint

72636c:
This is very MVPR but I suppose it closes #588, at least from an investigative standpoint.


samchungy:
🕺 💃 :D :D

72636c:
This has been collecting dust for too long and I think it would be more productive to chip away at the integration in smaller PRs against the default branch. It seems I've gotten far enough to get skuba to build itself, though there are a tonne of missing features and we probably need to clean up the error/warning output.

You're way more experienced with esbuild @samchungy and we'll likely lean on you heavily here 🙂.

src/global.d.ts
```diff
@@ -0,0 +1,9 @@
+// evanw/esbuild#2388
+
+declare global {
+  namespace WebAssembly {
+    interface Module {}
+  }
+}
```
72636c, line 7: Oops, forgot one thing: we need to figure out a way to include declaration in skuba's bundle and for TypeScript to pick it up in consuming repos, else you'll get compiler errors when type checking esbuild itself unless you have `DOM` types or `skipLibCheck: true`.

samchungy:
> This has been collecting dust for too long and I think it would be more productive to chip away at the integration in smaller PRs against the default branch. It seems I've gotten far enough to get skuba to build itself, though there are a tonne of missing features and we probably need to clean up the error/warning output.
> 
> You're way more experienced with esbuild @samchungy and we'll likely lean on you heavily here 🙂.

Tbh though I haven't used esbuild without `bundle` so I'd need a bit of time to 👀 at everything

src/cli/build/esbuild.ts
```diff
@@ -0,0 +1,126 @@
...
+    format: compilerOptions.module === ModuleKind.CommonJS ? 'cjs' : undefined,
+    outdir: compilerOptions.outDir,
+    logLevel: debug ? 'debug' : 'info',
+    logLimit: 0,
+    platform:
+      compilerOptions.moduleResolution === ModuleResolutionKind.NodeJs
+        ? 'node'
+        : undefined,
+    plugins: [
+      tsconfigPaths({
```
samchungy, line 104: Interesting - is the use of this plugin because of https://github.com/evanw/esbuild/issues/394?

samchungy (approved):

src/cli/build/esbuild.ts
72636c: Yep, spot on; I'll include this as a comment. Not super enthused by the v0 plugin but it's the simplest path for now; if we end up building our own plugins for more complex scenarios I reckon we could vendor/improve on this.

src/global.d.ts
72636c: I think 1494857 will do it, but this upcoming release probably necessitates a beta or two to ensure everything comes together correctly.

PR #690, opened 2021-11-22
Add `GitHub.putIssueComment` helper

72636c:

### DIFF
--- a/src/api/github/issueComment.ts
+++ b/src/api/github/issueComment.ts
@@ -4,20 +4,22 @@ import * as Git from '../git';
 
 import { getPullRequest } from './pullRequest';
 
-const getUserId = async (client: Octokit): Promise<string> => {
+const getUserId = async (client: Octokit): Promise<number> => {
   const { data } = await client.users.getAuthenticated();
 
-  return data.login;
+  return data.id;
 };
 
 /**
- * {@link https://docs.github.com/en/rest/reference/checks#create-a-check-run}
+ * https://docs.github.com/en/rest/reference/issues#create-an-issue-comment
  */
 interface PutIssueCommentParameters {
   /**
    * The body of the issue comment.
    */
   body: string;
+
+  env?: Record<string, string | undefined>;
 
   /**
    * The number that identifies the GitHub issue.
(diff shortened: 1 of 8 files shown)


src/api/github/issueComment.ts
```diff
@@ -0,0 +1,106 @@
...
+import { getPullRequest } from './pullRequest';
+
+const getUserId = async (client: Octokit): Promise<string> => {
+  const { data } = await client.users.getAuthenticated();
+
+  return data.login;
+};
+
+/**
+ * {@link https://docs.github.com/en/rest/reference/checks#create-a-check-run}
```
samchungy, line 14: link doesn't seem right fyi

src/api/github/issueComment.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  const dir = process.cwd();
+
+  const { owner, repo } = await Git.getOwnerAndRepo({ dir });
+
+  const client = new Octokit({ auth: apiTokenFromEnvironment() });
+
+  const issueNumber =
+    params.issueNumber ??
+    (await getPullRequest({ client, env }).then(({ number }) => number));
```
samchungy, line 81:
```suggestion
    (await getPullRequest({ client, env })).number;
```

src/api/github/pullRequest.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (!data.length) {
+    throw new Error(
+      `Commit ${commitId} is not associated with a GitHub pull request`,
+    );
+  }
+
+  data
+    .sort((a, b) => b.updated_at.localeCompare(a.updated_at))
+    .sort((a, b) => {
+      if (typeof a.closed_at === typeof b.closed_at) {
```
samchungy, line 71: question: Does mean the bot would update a comment on a closed PR if there are no open PRs? Is it worth it to update a comment on a closed PR?

72636c: Good catch, it probably makes sense to fail in this scenario.

src/api/github/pullRequest.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Gets the number of the current pull request.
+ *
+ * This tries to extract the pull request from common CI environment variables,
+ * and falls back to querying the GitHub Repos API for the latest pull request
+ * associated with the head commit. An error is thrown if there are no
+ * associated pull requests or if they are all closed.
+ */
+export const getPullRequest = async (
```
samchungy, line 31: question: Would it be worth exposing any of the other fields for the PR? I guess if I'm calling `getPullRequest` I would be expecting more than a simple number?

72636c: Yeah good point. I wanted to keep the API small until there was a need for extra fields, but I figure we can add a few more non-controversial ones.

72636c: On second thought, I'll scope this to `getPullRequestNumber`. The nicety here is that we can drive the function purely off of environment variables on the happy path.

samchungy (approved):

PR #695, opened 2021-11-25
Fix lambda-sqs-worker-cdk template

alazyzombie:
Did these things in https://github.com/SEEK-Jobs/asia-jobs-data-migrator in order to get it working:

1. Fixed the yarn test step in buildkite (it passes locally)
2. Fixed errors where node_modules weren't being included in the lambda

3. fixed the weird logs in cicd on deploying for cdk
image

### DIFF
--- a/.changeset/stupid-pets-explain.md
+++ b/.changeset/stupid-pets-explain.md
@@ -2,4 +2,4 @@
 "skuba": patch
 ---
 
-**template/lambda-sqs-worker-cdk:** Fix docker-compose volume mount and deploy output
+template/lambda-sqs-worker-cdk: Fix docker-compose volume mount and deploy output
--- a/.changeset/stupid-pets-explain.md
+++ b/.changeset/stupid-pets-explain.md
@@ -1,0 +1,5 @@
+---
+"skuba": patch
+---
+
+**template/lambda-sqs-worker-cdk:** Fix docker-compose volume mount and deploy output
--- a/template/lambda-sqs-worker-cdk/.buildkite/pipeline.yml
+++ b/template/lambda-sqs-worker-cdk/.buildkite/pipeline.yml
@@ -56,13 +56,12 @@ steps:
     label: 🧪 Test, Lint & Build
     artifact_paths: lib/**/*
     commands:
-      - yarn package
-      - echo '--- yarn build'
-      - yarn build
       - echo '+++ yarn test:ci'
       - yarn test
       - echo '--- yarn lint'
       - yarn lint
+      - echo '--- yarn build'
+      - yarn build
     env:
       GET_GITHUB_TOKEN: please
     plugins:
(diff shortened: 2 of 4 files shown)


template/lambda-sqs-worker-cdk/.buildkite/pipeline.yml
```diff
@@ -56,12 +56,13 @@ steps:
     label: 🧪 Test, Lint & Build
     artifact_paths: lib/**/*
     commands:
+      - yarn package
```
alazyzombie, line 59: doesn't feel right to have to run yarn package here when it's already in Dockerfile

samchungy: Ah this isn't working because we've forgotten to add the docker ecr cache plugins here. Can you try adding the following plugin section to the step?

```
  plugins:
        - *aws-sm
        - *private-npm
        - *docker-ecr-cache
        - docker-compose#v3.9.0:
            dependencies: false
            run: app
```

samchungy: Does your test step require you to run yarn build beforehand?

samchungy: Oh I see what might be happening here. Do we need to mount lib/node_modules as a volume too? @72636c

samchungy: Could you try something for me - could you add
`- /workdir/lib/node_modules` to the volumes section in your docker-compose file and undo all changes except the deploy step in your repo?

72636c: 💡 That sounds right - it looks like `lib/node_modules` from `yarn package` is never mounted.

TBH I'm questioning why `yarn package` is even in the Dockerfile, it sounds like it belongs on the deploy step. I believe that was the rationale behind it being `--offline`, ideally it would be runnable without requiring access to the npm read token.

samchungy: Ooo that might be due to yarn cache clean afaik we need the cache to run any with --offline

72636c: Curious, is the command doing anything that `cp -R` doesn't?

samchungy: Not sure aha, I think it may be possibly installing production node_modules in lib whilst leaving the Dev deps in tact? That's my theory

72636c: Ah of course, thanks

alazyzombie: this worked, thanks!

.changeset/stupid-pets-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+"skuba": patch
+---
+
+**template/lambda-sqs-worker-cdk:** Fix docker-compose volume mount and deploy output
```
72636c, line 5:
```suggestion
template/lambda-sqs-worker-cdk: Fix docker-compose volume mount and deploy output
```

72636c (approved):
🙇 Thanks a bunch!

PR #706, opened 2021-11-30
Add OpenTelemetry support for Koa template

runk:
This brings https://opentelemetry.io/ support, which can be enabled via gantry config (deploys another sidecar that collects the data). According to @bernos it's here to stay and should no longer be considered an experimental feature. Yet has to be widely announced. 

All traces can be viewed in "Technology Platforms" DD org.

This PR is a first step - one template only - koa app.

### DIFF
--- a/.changeset/brown-bananas-repair.md
+++ b/.changeset/brown-bananas-repair.md
@@ -2,4 +2,4 @@
 "skuba": minor
 ---
 
-template/koa-rest-api: Add opentelemetry support
+template/koa-rest-api: Add opt-in OpenTelemetry support
--- a/.changeset/brown-bananas-repair.md
+++ b/.changeset/brown-bananas-repair.md
@@ -1,0 +1,5 @@
+---
+"skuba": minor
+---
+
+template/koa-rest-api: Add opentelemetry support
--- a/template/koa-rest-api/Dockerfile
+++ b/template/koa-rest-api/Dockerfile
@@ -31,4 +31,4 @@ ARG PORT=8001
 ENV PORT ${PORT}
 EXPOSE ${PORT}
 
-CMD ["--require", "./lib/tracing.js", "lib/listen.js"]
+CMD ["--require", "./lib/tracing.js", "./lib/listen.js"]
--- a/template/koa-rest-api/Dockerfile
+++ b/template/koa-rest-api/Dockerfile
@@ -31,4 +31,4 @@ ARG PORT=8001
 ENV PORT ${PORT}
 EXPOSE ${PORT}
 
-CMD ["lib/listen.js"]
+CMD ["--require", "./lib/tracing.js", "lib/listen.js"]
--- a/template/koa-rest-api/gantry.apply.yml
+++ b/template/koa-rest-api/gantry.apply.yml
@@ -12,6 +12,7 @@ env:
 
   ENVIRONMENT: '{{.Environment}}'
   SERVICE: '{{values "service"}}'
+  OPENTELEMETRY_ENABLED: '{{.Values.openTelemetry.enabled | default false}}'
 
   {{range $key, $value := .Values.env}}
   {{$key}}: {{$value}}
@@ -21,6 +22,13 @@ env:
 datadogSecretId: '{{values "datadogSecretId"}}'
 {{end}}
 
+openTelemetry:
+  enabled: {{.Values.openTelemetry.enabled | default false}}
+  {{if .Values.isProduction}}
+  datadogEnvironmentName: production
+  {{else}}
+  datadogEnvironmentName: development
+  {{end}}
 {{if .Values.pagerDutyEndpoint}}
 pagerDutyEndpoint: '{{values "pagerDutyEndpoint"}}'
 {{end}}
--- a/template/koa-rest-api/gantry.apply.yml
+++ b/template/koa-rest-api/gantry.apply.yml
@@ -11,8 +11,8 @@ env:
   AWS_NODEJS_CONNECTION_REUSE_ENABLED: '1'
 
   ENVIRONMENT: '{{.Environment}}'
-  SERVICE: '{{values "service"}}'
   OPENTELEMETRY_ENABLED: '{{.Values.openTelemetry.enabled | default false}}'
+  SERVICE: '{{values "service"}}'
 
   {{range $key, $value := .Values.env}}
   {{$key}}: {{$value}}
--- a/template/koa-rest-api/package.json
+++ b/template/koa-rest-api/package.json
@@ -1,8 +1,13 @@
 {
   "dependencies": {
     "@koa/router": "^10.1.1",
+    "@opentelemetry/api": "^1.0.3",
+    "@opentelemetry/exporter-collector-grpc": "^0.25.0",
+    "@opentelemetry/instrumentation-aws-sdk": "^0.1.0",
+    "@opentelemetry/instrumentation-http": "^0.26.0",
+    "@opentelemetry/sdk-node": "^0.26.0",
     "@seek/logger": "^5.0.1",
-    "skuba-dive": "^1.2.0",
+    "aws-sdk": "^2.1039.0",
     "hot-shots": "^9.0.0",
     "koa": "^2.13.4",
     "koa-bodyparser": "^4.3.0",
@@ -11,6 +16,7 @@
     "runtypes-filter": "^0.6.0",
     "seek-datadog-custom-metrics": "^4.0.0",
     "seek-koala": "^5.1.0",
+    "skuba-dive": "^1.2.0",
     "uuid": "^8.3.2"
   },
   "devDependencies": {
--- a/template/koa-rest-api/src/listen.ts
+++ b/template/koa-rest-api/src/listen.ts
@@ -1,5 +1,3 @@
-import './register';
-
 import app from './app';
 import { config } from './config';
 import { rootLogger } from './framework/logging';
--- a/template/koa-rest-api/src/tracing.ts
+++ b/template/koa-rest-api/src/tracing.ts
@@ -38,7 +38,7 @@ const main = () => {
 
   sdk
     .start()
-    .then(() => log('info', 'Opentelemetry initialised'))
+    .then(() => log('info', 'OpenTelemetry initialised'))
     .catch((err: Error) =>
       log('error', 'Opentelemetry not initialised', { err }),
     );
--- a/template/koa-rest-api/src/tracing.ts
+++ b/template/koa-rest-api/src/tracing.ts
@@ -19,8 +19,8 @@ const log = (level: string, msg: string, extra = {}) => {
 };
 
 const main = () => {
-  // Use B3 propagation to ensure proper propagation between systems use OpenTelemetry and
-  // native Datadog APM, such as Istio/Envoy
+  // Use B3 propagation to ensure proper propagation between systems that use
+  // OpenTelemetry and native Datadog APM, such as Istio/Envoy.
   propagation.setGlobalPropagator(
     new CompositePropagator({
       propagators: [
--- a/template/koa-rest-api/src/tracing.ts
+++ b/template/koa-rest-api/src/tracing.ts
@@ -1,5 +1,5 @@
 /**
- * Opentelemetry tracing initialisation. This is a standalone ts/js module that is not
+ * OpenTelemetry tracing initialisation. This is a standalone TS/JS module that is not
  * referenced by application source code directly. It is required at runtime using the
  * node command's `--require` argument, see Dockerfile for details.
  */
--- a/template/koa-rest-api/src/tracing.ts
+++ b/template/koa-rest-api/src/tracing.ts
@@ -1,0 +1,61 @@
+/**
+ * Opentelemetry tracing initialisation. This is a standalone ts/js module that is not
+ * referenced by application source code directly. It is required at runtime using the
+ * node command's `--require` argument, see Dockerfile for details.
+ */
+
+import { propagation } from '@opentelemetry/api';
+import { CompositePropagator } from '@opentelemetry/core';
+import { CollectorTraceExporter } from '@opentelemetry/exporter-collector-grpc';
+import { AwsInstrumentation } from '@opentelemetry/instrumentation-aws-sdk';
+import { HttpInstrumentation } from '@opentelemetry/instrumentation-http';
+import { B3InjectEncoding, B3Propagator } from '@opentelemetry/propagator-b3';
+import { NodeSDK } from '@opentelemetry/sdk-node';
+
+const app = 'opentelemetry';
+const log = (level: string, msg: string, extra = {}) => {
+  const toLog = { msg, level, app, time: new Date().toISOString(), ...extra };
+  console.log(JSON.stringify(toLog)); // eslint-disable-line no-console
+};
+
+const main = () => {
+  // Use B3 propagation to ensure proper propagation between systems use OpenTelemetry and
+  // native Datadog APM, such as Istio/Envoy
+  propagation.setGlobalPropagator(
+    new CompositePropagator({
+      propagators: [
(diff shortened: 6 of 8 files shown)


template/koa-rest-api/src/tracing.ts
```diff
@@ -0,0 +1,45 @@
...
+  const sdk = new NodeSDK({
+    traceExporter: new CollectorTraceExporter(),
+    autoDetectResources: false,
+    instrumentations: [new HttpInstrumentation(), new AwsInstrumentation()],
+  });
+
+  return sdk.start();
+};
+
+const log = (level: string, msg: string, extra = {}) => {
```
runk, line 36: I'm hesitant to bring in the standard logger as it might bring more dependencies before internals instrimented. This lib really meant to be self-contained.

Thoughts?

template/koa-rest-api/src/listen.ts
```diff
@@ -1,5 +1,3 @@
-import './register';
```
runk, line 1: It's loaded in app.ts

template/koa-rest-api/gantry.apply.yml
```diff
@@ -21,6 +22,13 @@ env:
 datadogSecretId: '{{values "datadogSecretId"}}'
 {{end}}
 
+openTelemetry:
+  enabled: {{.Values.openTelemetry.enabled | default true}}
+  {{if .Values.isProduction}}
```
runk, line 27: There's a convention currently with this naming. Assuming it'd be great to keep it that way.

template/koa-rest-api/package.json
```diff
@@ -1,8 +1,13 @@
...
   "dependencies": {
     "@koa/router": "^10.1.1",
+    "@opentelemetry/api": "^1.0.3",
+    "@opentelemetry/exporter-collector-grpc": "0.25.0",
+    "@opentelemetry/instrumentation-aws-sdk": "0.1.0",
+    "@opentelemetry/instrumentation-http": "0.26.0",
+    "@opentelemetry/sdk-node": "0.26.0",
     "@seek/logger": "^5.0.1",
-    "skuba-dive": "^1.2.0",
+    "aws-sdk": "^2.1039.0",
```
samchungy, line 10: question: do you need to declare this as a dep?

template/koa-rest-api/package.json
```diff
@@ -1,8 +1,13 @@
 {
   "dependencies": {
     "@koa/router": "^10.1.1",
+    "@opentelemetry/api": "^1.0.3",
+    "@opentelemetry/exporter-collector-grpc": "0.25.0",
+    "@opentelemetry/instrumentation-aws-sdk": "0.1.0",
```
samchungy, line 6: question: should we be using ^ for the versions for consistency?

etaoins:
Has this been running in production yet? Any early performance insights versus no instrumentation or Datadog APM?

template/koa-rest-api/package.json
runk: I think it's up for discussion. It's needed for `aws-sdk` instrumentation, which I guess is highly desirable for anyone at seek. Perhaps some don't use `aws-sdk` but I bet it's the minority.

samchungy: I think when I flipped through the GitHub page it said you could use aws-sdk V3 as well?

runk:
> Has this been running in production yet? Any early performance insights versus no instrumentation or Datadog APM?

Otel is in production for about ~20 services today, however only some have code instrumentation enabled. Most of those 20 services have the flag set in gantry which enables tracing in auth-sidecar *only*. I can see there are only 4 services that have code instrumentation enabled across a couple of teams. There're some heavy-loaded systems like the API router that is instrumented though. Also it comes down that what we want to instrument, e.g. both `aws-sdk` and node's `http` or only one one of those.

Re DD APM - I think it'd be a similar performance hit as with automatic instrumentation as all outgoing requests (and other i/o commands) are wrapped. 

Stats for one of the services:

Screen Shot 2021-12-01 at 9 22 04 am

Screen Shot 2021-12-01 at 9 19 09 am

I'll try to get similar insights for another service that's hit a bit harder.

runk:
Okay, there're two discreet parts - app code instrumentation (controlled via `tracing.ts`) and auth sidecar instrumentation (controlled via gantry config). The previous chart was for the case when we already had Otel enabled in gantry (think auth sidecar) and just added code instrumentation in the app itself. No visible performance hit could be seen.

However, when enabled in gantry AND app, there's about 2-3ms latency increase. I'm inclined to believe it's mainly attributed to sidecar instrumentation.


Screen Shot 2021-12-01 at 9 53 15 am

etaoins:
Is that expected @bernos? I don't recall that showing up on `indie-graphql-server`. Seems like a large regression for a feature teams might not end up using.

etaoins (approved):
Talking with @runk offline, there's potentially a large performance regression for some workloads (33%-50%). I'm happy to include this as a default off to allow teams to opt-in to tracing if they find it useful. This is similar to what we do with Datadog APM.

runk:
Changed to be opt-in.

I think as an alternative (a middle ground?) we can to have the tracing on in auth sidecar by default, but keep app instrumentation off by default. I think this way possible performance hit as you experienced with DD APM will be an unlikely scenario.

runk:
Screen Shot 2021-12-01 at 11 10 07 am

The graph shows the app in three modes:

- Bare app (before 9:15)
- Sidecar + App instrumentation (9:15-10:38)
- Sidecar instrumentation only (10:38 onwards)

While latency increase is a single digit, it could mean a ~30% increase for a fast-responding app. Assuming if the app is doing a lot of network calls and all be instrumented (e.g. Experience APIs/Graphql servers), the more serious performance hit of app instrumentation should be expected. I do not see time spent for sidecar instrumentation to differ depending on nature of the app itself.

template/koa-rest-api/package.json
bernos: Actually, otel won't work with aws-sdk < V3. At least it wasn't in early testing. This was because in V2 of the aws SDK, request signing was happening _before_ otel header propagation. This meant that all instrumented requests to AWS services via the AWS SDK would fail on account of invalid signatures. This wasn't an issue in V3

template/koa-rest-api/src/tracing.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { propagation } from '@opentelemetry/api';
+import { CompositePropagator } from '@opentelemetry/core';
+import { CollectorTraceExporter } from '@opentelemetry/exporter-collector-grpc';
+import { AwsInstrumentation } from '@opentelemetry/instrumentation-aws-sdk';
+import { HttpInstrumentation } from '@opentelemetry/instrumentation-http';
+import { B3InjectEncoding, B3Propagator } from '@opentelemetry/propagator-b3';
+import { NodeSDK } from '@opentelemetry/sdk-node';
+
+const main = async () => {
```
bernos, line 15: Does `async` code loaded via `--require` behave any differently to `async` code called from the entry point? ie. is there a race here between instrumentation being initialised and the first requests being handed by the service?

bernos:
This is great, happy to roll it out, but please don't default any open telemetry options to `enabled` by default just yet. The cost implications of mass adoption are still not well understood, and we'd like to avoid any shock there. There are contingencies and compensations that we can make in the event that costs exceed the forecast, but we need a bit more control over the rate of adoption in order to have time to adapt if needs be. Also, I mentioned an incompatibility with aws-sdk < v3 that may have been addressed in newer releases of v2, but there should probably be a warning on the tin.

template/koa-rest-api/package.json
runk: @bernos re `aws-sdk` v2 - I can confirm instrumentation works for it with current otel libraries.

samchungy: Would it be better to just leave it out and let consumers byo sdk?

template/koa-rest-api/src/tracing.ts
runk: Good point. I think it's definitely a possibility that promises aren't fulfilled before the app loads. Today there's no way to pause for async execution in the bootstrap file supplied via `--require x.js`.

Really depends on what happens inside of the `sdk.start` fn: https://www.npmjs.com/package/@opentelemetry/sdk-node. Looking in code it only really returns a pending promise when we want auto resource detection which is off: https://github.com/seek-oss/skuba/pull/706/files#diff-4836bf38ef9493029c23deb46fdf06733bbc8b8cec520b9f538ea1a68cebf8eaR33

I'm going to remove async from the fn to keep it simple. However it's only hiding the possible problem with pending promise from `sdk.start()`

template/koa-rest-api/package.json
runk: It depends on your definition of "better", as always :). Do we want to bring in `aws-sdk` instrumentation for consumers out of the box? 

Screen Shot 2021-12-01 at 4 21 38 pm

Looking at this trace - I see clear advantages of knowing what the HTTP POST was actually is about, which in the case with `aws-sdk` instrumentation is very illustrative. With minimal built-in `http` instrumentation it'll be just another opaque HTTP request.

The problem is https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-sdk needs aws-sdk (or rather its types)

runk:
@samchungy @etaoins do you want to check it one more time before I merge? The notable change is the addition of process exit handling: https://github.com/seek-oss/skuba/pull/706/files#diff-4836bf38ef9493029c23deb46fdf06733bbc8b8cec520b9f538ea1a68cebf8eaR46

samchungy:
> @samchungy @etaoins do you want to check it one more time before I merge? The notable change is the addition of process exit handling: https://github.com/seek-oss/skuba/pull/706/files#diff-4836bf38ef9493029c23deb46fdf06733bbc8b8cec520b9f538ea1a68cebf8eaR46

@72636c your review probably holds the most weight here 😆

template/koa-rest-api/src/listen.ts
72636c: I think this is here to make it obvious (to humans but also to e.g. linters) that the import has side effects and must go first. It's easy to shoot yourself in the foot by importing something else before `./app`, which is part of the reason why I'd like to get rid of this hook soon™️.

template/koa-rest-api/package.json
```diff
@@ -1,8 +1,13 @@
 {
   "dependencies": {
     "@koa/router": "^10.1.1",
+    "@opentelemetry/api": "^1.0.3",
+    "@opentelemetry/exporter-collector-grpc": "^0.25.0",
+    "@opentelemetry/instrumentation-aws-sdk": "^0.1.0",
+    "@opentelemetry/instrumentation-http": "^0.26.0",
+    "@opentelemetry/sdk-node": "^0.26.0",
```
72636c, line 8: Do we need to be concerned about these v0s? Does OpenTelemetry promise some degree of stability?

template/koa-rest-api/Dockerfile
```diff
@@ -31,4 +31,4 @@ ARG PORT=8001
 ENV PORT ${PORT}
 EXPOSE ${PORT}
 
-CMD ["lib/listen.js"]
+CMD ["--require", "./lib/tracing.js", "lib/listen.js"]
```
72636c, line 34:
```suggestion
CMD ["--require", "lib/tracing.js", "lib/listen.js"]
```

Or the other way; I'm not fussed as long as we're consistent:

```suggestion
CMD ["--require", "./lib/tracing.js", "./lib/listen.js"]
```

.changeset/brown-bananas-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"skuba": minor
+---
+
+template/koa-rest-api: Add opentelemetry support
```
72636c, line 5:
```suggestion
template/koa-rest-api: Add opt-in OpenTelemetry support
```

Q: should we add this to the Express template too?

template/koa-rest-api/src/tracing.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * Opentelemetry tracing initialisation. This is a standalone ts/js module that is not
```
72636c, line 2:
```suggestion
 * OpenTelemetry tracing initialisation. This is a standalone TS/JS module that is not
```

template/koa-rest-api/src/tracing.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const sdk = new NodeSDK({
+    traceExporter: new CollectorTraceExporter(),
+    autoDetectResources: false,
+    instrumentations: [new HttpInstrumentation(), new AwsInstrumentation()],
+  });
+
+  sdk
+    .start()
+    .then(() => log('info', 'Opentelemetry initialised'))
```
72636c, line 41:
```suggestion
    .then(() => log('info', 'OpenTelemetry initialised'))
```

template/koa-rest-api/gantry.apply.yml
```diff
@@ -12,6 +12,7 @@ env:
 
   ENVIRONMENT: '{{.Environment}}'
   SERVICE: '{{values "service"}}'
+  OPENTELEMETRY_ENABLED: '{{.Values.openTelemetry.enabled | default false}}'
```
72636c, line 15:
```suggestion
  OPENTELEMETRY_ENABLED: '{{.Values.openTelemetry.enabled | default false}}'
  SERVICE: '{{values "service"}}'
```

template/koa-rest-api/src/tracing.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const app = 'opentelemetry';
+const log = (level: string, msg: string, extra = {}) => {
+  const toLog = { msg, level, app, time: new Date().toISOString(), ...extra };
+  console.log(JSON.stringify(toLog)); // eslint-disable-line no-console
+};
+
+const main = () => {
+  // Use B3 propagation to ensure proper propagation between systems use OpenTelemetry and
+  // native Datadog APM, such as Istio/Envoy
```
72636c, line 23:
```suggestion
  // Use B3 propagation to ensure proper propagation between systems that use
  // OpenTelemetry and native Datadog APM, such as Istio/Envoy.
```

.changeset/brown-bananas-repair.md
runk: Sure, do you want me to do it in the same PR or separate?

template/koa-rest-api/src/tracing.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { propagation } from '@opentelemetry/api';
+import { CompositePropagator } from '@opentelemetry/core';
+import { CollectorTraceExporter } from '@opentelemetry/exporter-collector-grpc';
+import { AwsInstrumentation } from '@opentelemetry/instrumentation-aws-sdk';
+import { HttpInstrumentation } from '@opentelemetry/instrumentation-http';
+import { B3InjectEncoding, B3Propagator } from '@opentelemetry/propagator-b3';
+import { NodeSDK } from '@opentelemetry/sdk-node';
+
+const app = 'opentelemetry';
```
72636c, line 15: For posterity, I was wondering where this `app` log field convention was from and it looks like our sidecar follows suit.

72636c (approved):

.changeset/brown-bananas-repair.md
72636c: Feel free to merge this one in first if that's easier

template/koa-rest-api/src/tracing.ts
etaoins: Old mate `ca-go-library` 🙈

`seek-auth-sidecar` was originally `ca-auth-sidecar` and used the default logger from `ca-go-library`. The newer `name` convention comes from Pino/Bunyan.

PR #714, opened 2021-12-29
Migrate to AWS CDK v2

72636c:
https://docs.aws.amazon.com/cdk/v2/guide/migrating-v2.html

Resolves #708.

### DIFF
--- a/template/lambda-sqs-worker-cdk/cdk.json
+++ b/template/lambda-sqs-worker-cdk/cdk.json
@@ -1,7 +1,6 @@
 {
   "app": "npx ts-node infra/index.ts",
   "context": {
-    "@aws-cdk/core:enableStackNameDuplicates": "true",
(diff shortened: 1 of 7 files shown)


template/lambda-sqs-worker-cdk/cdk.json
```diff
@@ -1,7 +1,6 @@
 {
   "app": "npx ts-node infra/index.ts",
   "context": {
-    "@aws-cdk/core:enableStackNameDuplicates": "true",
```
72636c, line 4: > Unsupported feature flag '@aws-cdk/core:enableStackNameDuplicates'. This flag existed on CDKv1 but has been removed in CDKv2. CDK will now behave as the same as when the flag is enabled.

raphsutti (approved):

PR #724, opened 2022-01-10
Revert to TypeScript 4.4

72636c:
We need to upgrade `eslint-config-seek` first and haven't had the chance to get around to that. This allows us to release #669 without having to deal with ESLint compatibility warnings.


src/cli/configure/processing/typescript.ts
```diff
@@ -64,7 +64,6 @@ const createImportFromExpression = (
           factory.createNamedImports(
             importNames.map((importName) =>
               factory.createImportSpecifier(
-                false,
```
72636c, line 67: Per https://github.com/seek-oss/skuba/pull/685/files#diff-6d9a6e21478a12b9e1ea38772d95bb567a69f65b55b3c4cd5a886c0ec3add710

samchungy, line 67: this seems suss

72636c: https://github.com/microsoft/TypeScript/wiki/API-Breaking-Changes#typescript-45

samchungy (approved):

PR #730, opened 2022-01-10
Restore Node.js 12 compatibility

72636c:
While Node.js 12 is going out of support soon, this was an unintended breaking change.


etaoins (approved):
🙇

samchungy (approved):

PR #754, opened 2022-02-09
Fix eskew typo in koa-rest-api readme

YesmynameisPerry:
Typo was introduced as part of #334, it was fixed in the express api template but not the koa one.


72636c (approved):
🙇

PR #767, opened 2022-02-11
Reduce Repetition of Environment variables

samchungy:
Use `provider.environment` to reduce repetition in the config


72636c (approved):
Oh, this is nice!

PR #775, opened 2022-02-20
template: Specify default Buildkite agent

72636c:
This is supported out of the box with newly created Buildkite pipelines using the new YAML Steps editor.

https://buildkite.com/docs/tutorials/pipeline-upgrade


template/greeter/.buildkite/pipeline.yml
```diff
@@ -1,11 +1,7 @@
-configs:
-  environments:
-    - &prod
-      agents:
-        queue: <%- prodBuildkiteQueueName %>
-      env:
-        ENVIRONMENT: prod
```
etaoins, line 7: We're okay losing the `ENVIRONMENT` here?

etaoins (approved):

template/greeter/.buildkite/pipeline.yml
72636c: Yeah, `greeter` is barebones so it doesn't control anything.

PR #785, opened 2022-03-10
Simplify Gantry API templates

bridgetbarnes:
The current API templates are geared towards making productionised system development very efficient, and are very good at it. 

However, there are use cases where one needs a service that can be built and deployed locally- workshops, for testing purposes, etc. This ~~adds a template that has a slightly paired down Dockerfile that makes running `gantry build` easier~~ modifies the koa rest template to have a Dockerfile that can be run using `gantry build` without modification


72636c:
CI is currently failing on npm/cli#4587 😞

docs/templates/api.md
```diff
@@ -24,4 +24,5 @@ A REST [resource API] built on the [Koa] web framework and deployed with [Gantry
 [express]: https://expressjs.com/
 [gantry]: https://gantry.ssod.skinfra.xyz
 [koa]: https://koajs.com/
-[resource api]: https://tech-strategy.ssod.skinfra.xyz/docs/v1/components.html#resource-api
+
+## [resource api]: https://tech-strategy.ssod.skinfra.xyz/docs/v1/components.html#resource-api
```
72636c:
```suggestion
[resource api]: https://tech-strategy.ssod.skinfra.xyz/docs/v1/components.html#resource-api
```

template/koa-rest-api/Dockerfile.dev-deps
```diff
@@ -1,11 +0,0 @@
-# syntax=docker/dockerfile:1.2
-
-FROM node:16-alpine AS dev-deps
-
-WORKDIR /workdir
-
-COPY package.json yarn.lock ./
-
-RUN \
-  --mount=type=secret,id=npm,dst=/workdir/.npmrc \
```
72636c, line 10: I think we need the `syntax` and `mount` in the `deps` stage of the main Dockerfile now, or otherwise we can drop the `private-npm` Buildkite plugin. (By default the template does not use any private SEEK packages, but the pipeline is wired up to support that once you need it.)

template/koa-rest-api/Dockerfile
```diff
@@ -1,15 +1,14 @@
...
-FROM ${BASE_IMAGE}:${BASE_TAG} AS deps
+COPY package.json yarn.lock ./
 
-RUN yarn install --ignore-optional --ignore-scripts --non-interactive --offline --production
+RUN yarn install --ignore-optional --ignore-scripts --non-interactive --production
 
 ###
 
-FROM ${BASE_IMAGE}:${BASE_TAG} AS build
+FROM deps AS build
```
72636c, line 11: This will need to be `FROM dev-deps as build` as we need `dev-deps` like skuba itself to be able to run the build (while prod `deps` only will be copied to the final runtime stage)

72636c:
Next steps should be:

1. Validate that these templates still deploy
2. Write the deep dive topic that explains how you to add ECR caching to the mix (I can pick this up)

template/express-rest-api/Dockerfile
```diff
@@ -1,15 +1,24 @@
...
+
+FROM node:16-alpine AS dev-deps
+
+WORKDIR /workdir
+
+COPY package.json yarn.lock ./
+
+RUN \
+  --mount=type=secret,id=npm,dst=/workdir/.npmrc \
+  yarn install --frozen-lockfile --ignore-optional --non-interactive
```
72636c, line 11: Oh, I just remembered why we were forced to go with two Dockerfiles: Gantry doesn't currently support build secrets, so it _can't_ run this `dev-deps` stage.

We could remove the `syntax` and `mount` instructions given the templates only use public packages, but it will mean that devs will invariably need to read our doco and split their Dockerfile once they need to consume private packages.

Alternatively, we could look into landing build secret support in Gantry first so we can land this PR as-is. I had a poke around today and it seems like we can achieve this by flipping `ImageBuild` into BuildKit mode, but it's definitely a non-trivial change.

72636c:
Temporarily closing this to reduce some Review Dino noise

PR #786, opened 2022-03-10
Suppress `eslint-plugin-react` warning

72636c:
We got bitten by seek-oss/eslint-config-seek#68.

Example: https://github.com/seek-oss/skuba/runs/5490668838?check_suite_focus=true#step:6:10

### DIFF
--- a/src/cli/adapter/eslint.ts
+++ b/src/cli/adapter/eslint.ts
@@ -54,6 +54,8 @@ export const runESLint = async (
       // `eslint-plugin-react` prints this annoying error on non-React repos.
       // We still want to support React linting for repos that have React code,
       // so we have to manually suppress it.
+      //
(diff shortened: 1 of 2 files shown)


src/cli/adapter/eslint.ts
```diff
@@ -46,11 +46,28 @@ export const runESLint = async (
...
   const start = process.hrtime.bigint();
 
+  /* eslint-disable no-console */
+  const ogConsoleError = console.error;
+  console.error = (...args: any[]) => {
+    if (
+      args[0] !==
+      // `eslint-plugin-react` prints this annoying error on non-React repos.
+      // We still want to support React linting for repos that have React code,
+      // so we have to manually suppress it.
```
72636c, line 56:
```suggestion
      // so we have to manually suppress it.
      //
      // https://github.com/yannickcr/eslint-plugin-react/blob/7484acaca8351a8568fa99344bc811c5cd8396bd/lib/util/version.js#L61-L65
```

raphsutti (approved):

PR #805, opened 2022-03-22
Pin GitHub action versions

72636c:


.github/workflows/codeql-analysis.yml
```diff
@@ -3,11 +3,10 @@
 #
 # You may wish to alter this file to override the set of languages analyzed,
 # or to provide custom queries or build logic.
-name: 'CodeQL'
+name: CodeQL
 
 on:
   push:
-    branches: [beta, main, master]
```
72636c, line 10: This should hopefully resolve the following warning:

> **Analyze (javascript)**
> 1 issue was detected with this workflow: Please make sure that every branch in on.pull_request is also in on.push so that Code Scanning can compare pull requests against the state of the base branch.

https://github.com/seek-oss/skuba/actions/runs/2020191009

etaoins (approved):

PR #807, opened 2022-03-22
Lock down GitHub workflows

72636c:
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions
- https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions


raphsutti (approved):
🔒

PR #829, opened 2022-04-07
Check for the "special pipeline name" as a substring of each splitted queuename

gitlon:
We have more than one central publishing pipeline - `artefacts:npm`, `artefacts:npm2` and possibly others in future?

Test for all of these when setting this flag, by changing this check to be a substring-search.

### DIFF
--- a/src/utils/args.ts
+++ b/src/utils/args.ts
@@ -10,7 +10,11 @@ export const hasSerialFlag = (args = process.argv, env = process.env) =>
   Boolean(
     // Run serially on SEEK's central npm publishing pipeline.
     // Exhausting agents here can cause grief.
-    env.BUILDKITE_AGENT_META_DATA_QUEUE?.split(',').includes('artefacts:npm'),
+    
+    env.BUILDKITE_AGENT_META_DATA_QUEUE?.split(',').reduce(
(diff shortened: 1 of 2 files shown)


src/utils/args.ts
```diff
@@ -10,7 +10,11 @@ export const hasSerialFlag = (args = process.argv, env = process.env) =>
   Boolean(
     // Run serially on SEEK's central npm publishing pipeline.
     // Exhausting agents here can cause grief.
-    env.BUILDKITE_AGENT_META_DATA_QUEUE?.split(',').includes('artefacts:npm'),
+    
+    env.BUILDKITE_AGENT_META_DATA_QUEUE?.split(',').reduce(
+      (accumulator, currentValue) => accumulator || currentValue.includes('artefacts:npm'),
+      false,
+    ),
```
72636c, line 17:
```suggestion
    env.BUILDKITE_AGENT_META_DATA_QUEUE?.split(',').some(
      (queueName) => queueName.startsWith('artefacts:npm')
    ),
```

72636c (approved):
Thanks for this! I hope I haven't inadvertently broken forked contributions 😬

PR #830, opened 2022-04-07
Try to run validate workflow on forks

72636c:
Secrets are not available to forks, so the empty `token` was failing in #829.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -24,10 +24,15 @@ jobs:
       checks: write
     runs-on: ubuntu-latest
     steps:
-      - name: Check out repo
+      - if: github.event.repository.fork == false
+        name: Check out repo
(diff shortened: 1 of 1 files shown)


samchungy (approved):

.github/workflows/validate.yml
```diff
@@ -24,11 +24,16 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - name: Check out repo
+      - if: github.event.repository.fork == false
+        name: Check out repo
         uses: actions/checkout@v3
         with:
           token: ${{ secrets.SEEK_OSS_CI_GITHUB_TOKEN }}
 
+      - if: github.event.repository.fork == true
```
72636c, line 33: This didn't work 😞, will have a poke later.

PR #842, opened 2022-04-14
Time out Buildkite test steps after 10 minutes

72636c:
Our projects should aim for a shorter duration for testing and linting. Taking longer than 10 minutes usually indicates a command or request has hung and will indefinitely hog the build agent, so we define a timeout.

### DIFF
--- a/.changeset/honest-apes-vanish.md
+++ b/.changeset/honest-apes-vanish.md
@@ -1,0 +1,13 @@
+---
+'skuba': patch
+---
+
+template: Time out Buildkite test steps after 10 minutes
+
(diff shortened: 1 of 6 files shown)


samchungy:
Could we add an auto retry for timeout failures?

72636c:
@samchungy have you observed many _transient_ timeouts? If so, that makes sense and I'll add 👍

72636c:
👌 I'd lean towards keeping this as-is then, we can always revisit in future. The most common issue I've seen is agent resource exhaustion, which is probably made worse by an automatic retry and better resolved by beefing up the instances.

samchungy (approved):

PR #843, opened 2022-04-14
Update build agent queue

phillip-le:
The "npm" agent queue is no longer updated. Instead, the "npm2" queue is used for new repositories. See https://github.com/SEEK-Jobs/gutenberg#installing-on-your-repository


72636c (approved):
I need to fix CI for forks 😬, but thanks for the contribution!

PR #845, opened 2022-04-16
Try to run validate workflow on forks deux

72636c:
Another attempt at #830.

https://github.community/t/is-there-a-way-to-tell-if-a-pr-is-from-a-forked-repository/134186

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -24,15 +24,20 @@ jobs:
       checks: write
     runs-on: ubuntu-latest
     steps:
-      - if: github.event.repository.fork == false
+      - if: >-
+          github.event_name != 'pull_request' ||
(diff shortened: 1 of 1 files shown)


.github/workflows/validate.yml
```diff
@@ -24,15 +24,20 @@ jobs:
       checks: write
     runs-on: ubuntu-latest
     steps:
-      - if: github.event.repository.fork == false
+      - if: >-
```
samchungy, line 27: What does the `>-` do?

samchungy (approved):

.github/workflows/validate.yml
72636c: It's a header for collapsing the multiline string: https://yaml-multiline.info/

PR #864, opened 2022-05-14
feat: use `AsyncLocalStorage` to track logger context

samchungy:
This might be a little radical but overall improves the DX and allows users to not need to worry about passing a context object around and/or a context logger around the codebase.

Some local testing confirms `x-request-id` is still present:

image

Performance hit of ~1.75% according to this which is negligible

https://github.com/nodejs/node/issues/34493#issuecomment-845094849

### DIFF
--- a/.changeset/olive-worms-run.md
+++ b/.changeset/olive-worms-run.md
@@ -4,10 +4,10 @@
 
 template/koa-rest-api: Use [AsyncLocalStorage](https://nodejs.org/docs/latest-v16.x/api/async_context.html#asynchronous-context-tracking) to track logger context
 
-
 We now employ [RequestLogging.createContextStorage](https://github.com/seek-oss/koala/blob/master/src/requestLogging/README.md#context-logging) to thread logging context through the middleware stack of your Koa application. This enables use of a singleton `logger` instance instead of manually propagating Koa context and juggling `rootLogger`s and `contextLogger`s.
 
 Before:
+
 ```typescript
 import createLogger from '@seek/logger';
 import Koa from 'koa';
@@ -20,7 +20,9 @@ const app = new Koa().use((ctx) => {
   contextLogger(ctx).info('Has context');
 });
 ```
+
 After:
+
 ```typescript
 import createLogger from '@seek/logger';
 import Koa from 'koa';
--- a/.changeset/olive-worms-run.md
+++ b/.changeset/olive-worms-run.md
@@ -1,0 +1,5 @@
+---
+'skuba': patch
+---
+
+template/koa-rest-api: Use AsyncLocalStorage to track logger context
--- a/.changeset/olive-worms-run.md
+++ b/.changeset/olive-worms-run.md
@@ -2,4 +2,34 @@
 'skuba': patch
 ---
 
-template/koa-rest-api: Use `createContextStorage` from seek-koala to track logger context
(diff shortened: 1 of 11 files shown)


.changeset/olive-worms-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+template/koa-rest-api: Use AsyncLocalStorage to track logger context
```
72636c, line 5:
```suggestion
template/koa-rest-api: Use [AsyncLocalStorage](https://nodejs.org/docs/latest-v16.x/api/async_context.html#asynchronous-context-tracking) to track logger context
```

72636c (approved):
This looks excellent and it sounds like there is little performance penalty with Node.js 16: https://github.com/nodejs/node/issues/34493#issuecomment-845094849. I think our Lambdas could similarly benefit from the same approach.

It's pretty trivial but I wonder if we should upstream something to https://github.com/seek-oss/koala to enforce this pattern.

.changeset/olive-worms-run.md
samchungy: > This looks excellent and it sounds like there is little performance penalty with Node.js 16: [nodejs/node#34493 (comment)](https://github.com/nodejs/node/issues/34493#issuecomment-845094849). I think our Lambdas could similarly benefit from the same approach.
> 
> It's pretty trivial but I wonder if we should upstream something to https://github.com/seek-oss/koala to enforce this pattern.

Ooooo that might be an idea. We could potentially initialise it via seek koala and ship the middleware from there too

samchungy:
Going to wait for the koala PR to go through before changing this to use that implementation.

.changeset/olive-worms-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+template/koa-rest-api: Use [AsyncLocalStorage](https://nodejs.org/docs/latest-v16.x/api/async_context.html#asynchronous-context-tracking) to track logger context
```
72636c:
````suggestion
template/koa-rest-api: Use [AsyncLocalStorage](https://nodejs.org/docs/latest-v16.x/api/async_context.html#asynchronous-context-tracking) to track logger context


We now employ [RequestLogging.createContextStorage](https://github.com/seek-oss/koala/blob/master/src/requestLogging/README.md#context-logging) to thread logging context through the middleware stack of your Koa application. This enables use of a singleton `logger` instance instead of manually propagating Koa context and juggling `rootLogger`s and `contextLogger`s.

Before:

```typescript
import createLogger from '@seek/logger';
import Koa from 'koa';
import { RequestLogging } from 'seek-koala';

const rootLogger = createLogger();

const contextLogger = (ctx: Context) =>
  rootLogger.child(RequestLogging.contextFields(ctx));

const app = new Koa().use((ctx) => {
  rootLogger.info('Has no context');

  contextLogger(ctx).info('Has context');
});
```

After:

```typescript
import createLogger from '@seek/logger';
import Koa from 'koa';
import { RequestLogging } from 'seek-koala';

const { createContextMiddleware, mixin } =
  RequestLogging.createContextStorage();

const contextMiddleware = createContextMiddleware();

const logger = createLogger({ mixin });

const app = new Koa().use(contextMiddleware).use((ctx) => {
  logger.info('Has context');
});
```
````

72636c (approved):
🚢

PR #870, opened 2022-05-20
Clear timeout

72636c:
This could cause the process to hang around until the timeout resolved.


samchungy:
Would it be worth using the [timers promise api ](https://nodejs.org/api/timers.html#timers-promises-api) given that its now stable in node 16?

samchungy (approved):

72636c:
@samchungy That would be much nicer, but I don't want to bump the minimum version just yet as it would break install on most existing Lambda projects. Ideally we'd implement some Node.js upgrade autofix/command first.

PR #879, opened 2022-05-26
Fix Context Logger

samchungy:
https://seekchat.slack.com/archives/CDCPCEPV3/p1653606310727419

### DIFF
--- a/.changeset/cool-buses-give.md
+++ b/.changeset/cool-buses-give.md
@@ -1,0 +1,5 @@
+---
+'skuba': patch
+---
+
+template/lambda-sqs-worker: Fix a bug where the logger was mutating the context. We now perform a shallow copy when retrieving the logger context from AsyncLocalStorage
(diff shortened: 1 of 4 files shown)


.changeset/large-swans-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+template/koa-rest-api: Update `seek-koala` to prevent logger context mutation
```
72636c, line 5: We should be able to do without this Changeset because it's a caret range; any `init`s from here will pick up the newest version.

.changeset/cool-buses-give.md
```diff
@@ -0,0 +1,10 @@
+---
+'skuba': patch
+---
+
+template/lambda-sqs-worker: Fixed a bug where the logger was mutating the context. We now perform a shallow copy when retrieving the logger context from AsyncLocalStorage
```
72636c, line 5:
```suggestion
template/lambda-sqs-worker: Avoid mutation of logger context

We now perform a shallow copy when retrieving the logger context from `AsyncLocalStorage`.
```

72636c (approved):

PR #908, opened 2022-06-27
Remove Unnecessary Serverless Permission

samchungy:
This perm is only needed if we configure asynchronous invocation and use the `onError` configuration but SQS is a synchronous invocation and SQS handles the dead-lettering itself.

### DIFF
--- a/.changeset/sharp-ads-clean.md
+++ b/.changeset/sharp-ads-clean.md
@@ -1,5 +1,5 @@
 ---
-"skuba": patch
+'skuba': patch
 ---
 
 template/lambda-sqs-worker: Remove unnecessary IAM permission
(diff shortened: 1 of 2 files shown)


72636c (approved):
Good catch. This was a copy+paste from our internal guidance for workers to perform their own in-process re-queuing and DLQing to avoid throttling, though without the actual logic to do so because we haven't open sourced any helper package to achieve this yet.

https://tech-strategy.ssod.skinfra.xyz/docs/v1/technology.html#performance-degradation-on-lambda-function-errors

This has been partially superseded by batch failure reporting, though the "immediate DLQ" case still stands:

https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting

PR #940, opened 2022-08-15
make linter ignore cdk.out directory

dan753722:
Add ignore cdk.out directory linting rule.

`yarn format` or `yarn lint` will probably trigger node heap out of memory issues otherwise.

### DIFF
--- a/template/base/_.prettierignore
+++ b/template/base/_.prettierignore
@@ -3,6 +3,7 @@
 .serverless/
 .vscode/
 node_modules*/
+cdk.out/
 
 /coverage*/
 /dist*/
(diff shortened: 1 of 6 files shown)


template/base/_.prettierignore
```diff
@@ -14,4 +14,5 @@ node_modules*/
 /.gantry/**/*.yml
 gantry*.yaml
 gantry*.yml
+cdk.out*/
```
72636c, line 17: 1. Can you avoid smushing this together with the Gantry ignores? It doesn't not seem related and is probably more at home alongside the first block with `node_modules` and such.
2. Is the asterisk necessary?

72636c (approved):

PR #951, opened 2022-08-25
Sort `package.json` based on the well-known package.json keys

### LINKED ISSUE #926: `skuba configure` sorts differently to `yarn`
image

Seems to be an issue between the ordering of `_` and `-`. Causes these sort of commits to pop up in our PRs.

Raising an issue from our convo in case someone runs into this.

mrm007:
This uses [sort-package-json](https://github.com/keithamus/sort-package-json).

> ### How does it sort?
>
> It sorts using [`sort-object-keys`](http://github.com/keithamus/sort-object-keys). It sorts using the well-known keys of a package.json. For the full list check the [default rules](./defaultRules.md). It sorts sub-keys too - sometimes by a well-known order, other times alphabetically. The initial order was derived from the [package.json docs](https://docs.npmjs.com/files/package.json) with a few extras added for good measure.

Re: #107, sorting the `scripts` key [can also be enabled](https://github.com/keithamus/sort-package-json/blob/master/defaultRules.md#scripts). It says only the scripts in that list are sorted, but I tested with non-standard ones and it still worked — see latest commit 0cd952110ec9b7845116aedfd4c74d8ed1050ac4.

> Keys are sorted alphabetically except for [pre/post scripts](https://docs.npmjs.com/cli/v6/using-npm/scripts#pre--post-scripts). Those are placed before and after their corresponding base npm script.

Also fixes #926.

### DIFF
--- a/.changeset/twenty-pigs-sparkle.md
+++ b/.changeset/twenty-pigs-sparkle.md
@@ -1,0 +1,5 @@
+---
+'skuba': minor
+---
+
+**configure, init:** Format `package.json` with [sort-package-json](https://github.com/keithamus/sort-package-json)
--- a/.changeset/twenty-pigs-sparkle.md
+++ b/.changeset/twenty-pigs-sparkle.md
@@ -2,4 +2,4 @@
 'skuba': minor
 ---
 
-**configure, init:** Format `package.json` with [sort-package-json](https://github.com/keithamus/sort-package-json)
+configure, init: Format `package.json` with [sort-package-json](https://github.com/keithamus/sort-package-json)
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "semantic-release": "^19.0.0",
     "serialize-error": "^8.0.1",
     "simple-git": "^3.5.0",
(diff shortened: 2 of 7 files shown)


.changeset/twenty-pigs-sparkle.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': minor
+---
+
+**configure, init:** Format `package.json` with [sort-package-json](https://github.com/keithamus/sort-package-json)
```
72636c, line 5: I think this will be bolded automatically: https://github.com/seek-oss/skuba/blob/3d9fa47704bb1a6e8397347c46f1cc27b3b6b709/.changeset/changelog.js#L13

```suggestion
configure, init: Format `package.json` with [sort-package-json](https://github.com/keithamus/sort-package-json)
```

72636c (approved):
This looks amazing! Does it also close #926?

mrm007:
@72636c yep — bafed3f41e3ad6e125d38d19722717914ea35818

samchungy:
Nice one 🥳

PR #986, opened 2022-10-09
[WIP] Default to ESM Package Release

samchungy:
https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing

This should still be CJS backwards compatible. 🤔 Still need to test this more


template/oss-npm-package/_package.json
```diff
@@ -6,6 +6,13 @@
     "commitizen": "^4.2.4",
     "skuba": "*"
   },
+  "exports": {
+    ".": {
+      "types": "./lib-types/index.d.ts",
+      "import": "./lib-es2015/index.js",
+      "require": "./lib-commonjs/index.js"
```
72636c, line 13: Likely way beyond the scope of this PR, but just spitballing: I'm assuming that the most friendly hybrid approach going forwards would be a single `lib` folder with `index.d.ts`, `index.cjs` and `index.mjs` all at the same level. Is that worth exploring, and would we try to support that via esbuild/tsc/both?

samchungy: Not sure if I can even go ahead with this PR. I gave this a shot and looks like ts-node doesn't support it. As soon as it sees "type": "module" it gave up :(

PR #987, opened 2022-10-10
Enable isolated modules by default

### LINKED ISSUE #626: Make Jest go faster
As discussed with @samchungy, we can look to:

1. Default ts-jest's `isolatedModules` option to `true` in our Jest preset: https://kulshekhar.github.io/ts-jest/docs/getting-started/options/isolatedModules

   We already type check in the `skuba lint` command so doing it again on `skuba test` is unnecessary.

2. Investigate alternative transformers like https://github.com/aelbore/esbuild-jest
3. Default templates back to vanilla `jest.config.js` if the config transpile is noticeably slower

72636c:
Closes #626.


jest/moduleNameMapper.js
```diff
@@ -1,86 +1,58 @@
...
-const DEFAULT_PATHS = { src: ['src'], 'src/*': ['src/*'] };
+const { tryParseTsConfig } = require('./tsConfig');
 
-/**
- * @returns {unknown}
- */
-const getConfigFromDisk = () => {
-  const filename =
-    findConfigFile('.', sys.fileExists.bind(this)) ?? 'tsconfig.json';
+module.exports.createModuleNameMapper = (getConfig) => {
```
72636c, line 7: Sorry for the big diff here, functional changes should be:

1. Moving the try-catch to shared function in `jest/tsConfig.js`
2. Defaulting to no paths rather than `src`

samchungy (approved):

Issue #968, opened 2022-09-10
Pare down Koa template for local use


72636c:
As raised by @bridgetbarnes, the present `koa-rest-api` template is a bit bloated for a starting point and its Docker/Gantry configuration is difficult to run and deploy locally.

This is an issue to track SEEK-Jobs/gantry#1702; once resolved, we can restore #785.

PR #1021, opened 2022-11-07
Support AMD64 Docker builds via `BUILDPLATFORM`

72636c:
Thoughts welcome.

While we don't make it easy to build our templated Dockerfiles locally (at least until we can land #968), but we currently punish x86 holdouts by requiring them to emulate ARM. This should still default to ARM on the ARM64 build agents and deployment targets that we prescribe but allows for flexibility via the `BUILDPLATFORM` Docker built-in.

https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images

### DIFF
--- a/.changeset/thick-ducks-relax.md
+++ b/.changeset/thick-ducks-relax.md
@@ -1,0 +1,7 @@
+---
+'skuba': patch
+---
+
+template: Support AMD64 Docker builds via `BUILDPLATFORM`
+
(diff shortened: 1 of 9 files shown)


samchungy (approved):

samchungy:
Can you specify buildplatform via bk? I think last time I tried I wasn't having fun with the plugins.

72636c:
@samchungy BUILDPLATFORM _should_ be defaulted based on the instance type. Were you seeing something different?

72636c:
Ah right, `--platform` defaults to `--platform=${TARGETPLATFORM}` and I think we need to use `buildx` to access that properly.

PR #1025, opened 2022-11-08
Define BUILDPLATFORM fallback for API templates

72636c:
I was hoping that this global would always be available, but it looks like Gantry does not currently supply it when building images.

### DIFF
--- a/docs/deep-dives/arm64.md
+++ b/docs/deep-dives/arm64.md
@@ -50,12 +50,15 @@ See [Builds at SEEK] and the [Gantry ARM reference] for more information.
 If you are not ready to build and run your projects on ARM64 hardware,
 you can downgrade a templated project to be compatible with AMD64 hardware.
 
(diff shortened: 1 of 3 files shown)


samchungy (approved):
Would we need to change the gantry.build platform value too? My other thought was in the template initialiser we could get them to specify arm64 or amd64?

https://github.com/seek-oss/skuba/blob/fb8699fe237d86f78dc79f8f368d02bd55543f11/template/koa-rest-api/gantry.build.yml#L15

72636c:
@samchungy I don't think we can rely on `BUILDPLATFORM` in `gantry.build.yml` unless/until we replicate that Docker functionality in Gantry. We do mention it as a manual fix-up later in the doco: https://github.com/seek-oss/skuba/blob/0b1a9a3aff77972615305040844527438bd96217/docs/deep-dives/arm64.md?plain=1#L64-L69

But I think a templating option makes sense actually, I'll look at that 👍

PR #1036, opened 2022-11-10
Fix Jest Transformer

samchungy:
Transform config can also come in an array 🤷‍♂️

### DIFF
--- a/jest/transform.js
+++ b/jest/transform.js
@@ -31,22 +31,20 @@ module.exports.transform = Object.fromEntries(
       ];
     }
 
-    if (Array.isArray(value)) {
-      return [
-        key,
-        value[0] === TS_JEST_NAME
-          ? [
-              TS_JEST_PATH,
-              {
-                ...value[1],
-                isolatedModules:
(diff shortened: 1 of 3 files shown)


jest/transform.js
```diff
@@ -18,7 +18,8 @@ const maybeTsConfig = tryParseTsConfig();
 module.exports.transform = Object.fromEntries(
   Object.entries(defaults.transform).map(([key, value]) => [
     key,
-    value === TS_JEST_NAME
+    (typeof value === 'string' && value === TS_JEST_NAME) ||
+    (Array.isArray(value) && value[0] === TS_JEST_NAME)
```
72636c, line 22: Should we do something like this for the array scenario:

```typescript
[
  TS_JEST_PATH,
  { ...value[1], isolatedModules: value[1].isolatedModules ?? maybeTsConfig.options.isolatedModules ?? true }
]
```

samchungy: Good call, I've split up the different types since it was getting messy

72636c (approved):
🙇

PR #1057, opened 2022-11-24
feat: used toHaveBeenCalledWith

lukeymyuan:
Replaced `mock.calls` with `toHaveBeenCalledWith` which is cleaner and more readable 

Co-authored-by: Conrad Lang <ConradLang@users.noreply.github.com>


samchungy (approved):

72636c:
Note that these are not equivalent. The previous code is exhaustive in asserting the total number of calls the logger receives, while the proposed code would allow additional logs to be emitted.

samchungy:
> Note that these are not equivalent. The previous code is exhaustive in asserting the total number of calls the logger receives, while the proposed code would allow additional logs to be emitted.

Ah good point. I guess the equivalent would be toBeCalledTimes(2) but at that point not sure which is more cleaner. 3 v 1 expects

lukeymyuan:
yeah.. i also thought about it...
let's just apply the changes where theres only 1 call

template/lambda-sqs-worker/src/framework/handler.test.ts
```diff
@@ -40,7 +40,7 @@ describe('createHandler', () => {
 
     await expect(handler(input, ctx)).rejects.toThrow('Function failed');
 
-    expect(logger.error.mock.calls).toEqual([[{ err }, 'Function failed']]);
+    expect(logger.error).toHaveBeenCalledWith({ err }, 'Function failed');
```
trevorgk, line 43: Just a thought...
```suggestion
    expect(logger.error).toHaveBeenCalledTimes(1);
    expect(logger.error).toHaveBeenCalledWith({ err }, 'Function failed');
```

72636c (approved):
Thanks, and sorry for the delay on this one! I think the proposed changes make sense; we care more that _an_ error log was written in these scenarios, not that it was the _only_ thing written.

PR #1068, opened 2022-12-19
Sign commits on GH Actions using GH API

samchungy:
Given the push for signed commits, set Skuba to use signed commits on GH Actions.


src/cli/lint/autofix.ts
```diff
@@ -95,22 +93,6 @@ export const autofix = async (params: AutofixParameters): Promise<void> => {
...
-      if (!ref) {
-        return log.warn('No autofixes detected.');
-      }
-
-      await throwOnTimeout(simpleGit().push(), { s: 30 });
-      log.warn(`Pushed fix commit ${ref}.`);
-      return;
-    }
-
     // Other CI Environments, use GitHub API
```
72636c, line 96:
```suggestion
```

72636c (approved):
Makes sense. Do you know if:

1. This documentation is still current and default actions `GITHUB_TOKEN` is insufficient to run autofixes: https://seek-oss.github.io/skuba/docs/deep-dives/github.html#github-autofixes
2. Which permissions are required for the PAT so we can document them (and ensure that `seek-oss` is configured appropriately)?

72636c:
Closing this for now, let's reopen when we have time to revisit 🙏

PR #1074, opened 2022-12-29
Exclude internal files from autofix commits

72636c:
Follows on from #1072 and additionally prevents `.npmrc` mishaps.

### DIFF
--- a/.changeset/red-llamas-explain.md
+++ b/.changeset/red-llamas-explain.md
@@ -1,0 +1,10 @@
+---
+'skuba': patch
+---
+
+lint: Exclude internal files from autofix commits
+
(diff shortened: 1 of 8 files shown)


samchungy:
should we try and read the .gitignore for max safety?

72636c:
@samchungy These code paths use `Git.getChangedFiles` under the hood, and from local testing it appears to respect `.gitignore` already 😌. This is an additional layer of protection as I think there are some projects where `.npmrc` is not `.gitignore`d; we have one `pnpm` repo where this is the case.

I was actually thinking of rolling back #1072 if we're happy with this approach to avoid the unnecessary `.gitignore` change.

etaoins (approved):

PR #1099, opened 2023-02-03
Update IDE advice for Webstorm

samchungy:
As much as vscode is the goat 🐐 . We also need to provide support to the plebs


72636c (approved):
I love clicking around GUIs

PR #1101, opened 2023-02-05
Add `skuba start` fastify support

samchungy:
Tested on my upcoming template:

https://github.com/samchungy/test-fastify-template/pull/1

### DIFF
--- a/src/wrapper/requestListener.ts
+++ b/src/wrapper/requestListener.ts
@@ -18,7 +18,7 @@ interface ObjectConfig {
   // Fastify compatibility
   server?: http.Server;
 
-  default?: unknown;
+  default?: Promise<unknown>;
   port?: unknown;
 }
 
@@ -46,7 +46,7 @@ export const runRequestListener = async ({
     return;
   }
 
-  let config = await entryPoint;
+  let config: FunctionConfig | ObjectConfig = await entryPoint;
 
   if (typeof config === 'object' && isConfig(config.default)) {
     // Prefer `export default` over `export =`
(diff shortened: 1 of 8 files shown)


src/wrapper/requestListener.ts
```diff
@@ -15,11 +15,16 @@ interface ObjectConfig {
...
+  server?: http.Server;
+
   default?: unknown;
   port?: unknown;
 }
 
-const isConfig = (data: unknown): data is FunctionConfig | ObjectConfig =>
+const isConfig = (
+  data: unknown,
+): data is Promise<FunctionConfig> | Promise<ObjectConfig> =>
```
72636c, line 27: 🤔 Why are these promises?

72636c (approved):

src/wrapper/requestListener.ts
samchungy: Fastify is a confusing hybrid of callbacks and promises. To support both styles we gotta make it a Promise.

https://www.fastify.io/docs/latest/Guides/Getting-Started/#your-first-server

samchungy: image

72636c: I don't think this guard is properly checking for a nested `Promise` type though. The input would have to be awaited before being passed in

samchungy: https://github.com/samchungy/test-fastify-template/pull/1 if you wanna peek. Still very much a WIP. Putting down the tools for the day :)

samchungy: The app.ts calls createApp() and exports it so it should be okay

samchungy: Alright added some tests and better typings

72636c (approved):
This looks great, thanks!

Up to you on whether you want to changeset this PR or leave it until you have more to share 🙂

PR #1111, opened 2023-03-02
Set the keep alive timeout to 31 seconds

samchungy:
The default Node.js server keep alive timeout is set to 5 seconds. However, the Gantry default ALB idle timeout is 30 seconds. This would lead to the occasional issues where the sidecar would throw `proxyStatus` 502 errors while trying to talk to your node application as it would believe that the socket is still open. AWS recommends setting an application timeout larger than the ALB idle timeout.

A more detailed explanation can be found in the below links

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
https://nodejs.org/docs/latest-v18.x/api/http.html#serverkeepalivetimeout

So far it has completely removed all 502 errors from the products-api and hirer graphql api

### DIFF
--- a/.changeset/fair-shrimps-count.md
+++ b/.changeset/fair-shrimps-count.md
@@ -2,11 +2,11 @@
 'skuba': patch
 ---
 
-template/\*-rest-api: Set `keepAliveTimeout` to 31 seconds to prevent 502 errors
+template/\*-rest-api: Set `keepAliveTimeout` to 31 seconds to prevent HTTP 502s
 
-The default Node.js server keep alive timeout is set to 5 seconds. However, the Gantry default ALB idle timeout is 30 seconds. This would lead to the occasional issues where the sidecar would throw `proxyStatus` 502 errors. AWS recommends setting an application timeout larger than the ALB idle timeout.
+The default Node.js server keep-alive timeout is set to 5 seconds. However, the Gantry default ALB idle timeout is 30 seconds. This would lead to the occasional issues where the sidecar would throw `proxyStatus=502` errors. AWS recommends setting an application timeout larger than the ALB idle timeout.
 
-A more detailed explanation can be found in the below links
+A more detailed explanation can be found in the below links:
 
-https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
-https://nodejs.org/docs/latest-v18.x/api/http.html#serverkeepalivetimeout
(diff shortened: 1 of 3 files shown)


.changeset/fair-shrimps-count.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+template/\*-rest-api: Set `keepAliveTimeout` to 31 seconds to prevent 502 errors
+
+The default Node.js server keep alive timeout is set to 5 seconds. However, the Gantry default ALB idle timeout is 30 seconds. This would lead to the occasional issues where the sidecar would throw `proxyStatus` 502 errors. AWS recommends setting an application timeout larger than the ALB idle timeout.
+
+A more detailed explanation can be found in the below links
+
+https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
+https://nodejs.org/docs/latest-v18.x/api/http.html#serverkeepalivetimeout
```
72636c, line 12:
```suggestion
template/\*-rest-api: Set `keepAliveTimeout` to 31 seconds to prevent HTTP 502s

The default Node.js server keep-alive timeout is set to 5 seconds. However, the Gantry default ALB idle timeout is 30 seconds. This would lead to the occasional issues where the sidecar would throw `proxyStatus=502` errors. AWS recommends setting an application timeout larger than the ALB idle timeout.

A more detailed explanation can be found in the below links:

1. <https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout>
2. <https://nodejs.org/docs/latest-v18.x/api/http.html#serverkeepalivetimeout>
```

template/express-rest-api/src/listen.ts
```diff
@@ -14,3 +14,10 @@ const listener = app.listen(config.port, () => {
     rootLogger.debug(`listening on port ${address.port}`);
   }
 });
+
+// Gantry ALB Default Idle timeout is 30 seconds
```
72636c, line 18:
```suggestion
// Gantry ALB default idle timeout is 30 seconds
```

template/koa-rest-api/src/listen.ts
```diff
@@ -15,3 +15,10 @@ const listener = app.listen(config.port, () => {
     logger.debug(`listening on port ${address.port}`);
   }
 });
+
+// Gantry ALB Default Idle timeout is 30 seconds
```
72636c, line 19:
```suggestion
// Gantry ALB default idle timeout is 30 seconds
```

72636c (approved):
🙇

PR #1122, opened 2023-03-18
Tweak changesets

72636c:


samchungy (approved):

.changeset/ninety-geckos-perform.md
```diff
@@ -1,7 +1,7 @@
 ---
-'skuba': patch
+'skuba': minor
 ---
 
 lint: Commit codegen updates
 
-`skuba lint` can locally codegen updates to ignore files and module exports. These changes are now automatically committed if you have [GitHub autofixes](https://seek-oss.github.io/skuba/docs/deep-dives/github.html#github-autofixes) enabled on your project.
+`skuba lint` can locally codegen updates to ignore files, module exports and Renovate configuration. These changes are now automatically committed if you have [GitHub autofixes](https://seek-oss.github.io/skuba/docs/deep-dives/github.html#github-autofixes) enabled on your project.
```
samchungy, line 7: Oooo does this mean I can push my generated openapi schema if I generate it before skuba lint?

72636c: It's currently limited to a hardcoded set of filenames that skuba may conceivably modify, due to paranoia around committing private ephemeral files like an `.npmrc`: https://github.com/seek-oss/skuba/blob/6270b786aea0fddeb77a88271295266df059636a/src/cli/lint/autofix.ts#L121-L124

However, I think it would be reasonable to extend this logic to include well-known and insensitive paths/files like your OpenAPI schema.

samchungy: Gotcha. Probably could make that configurable via maybe the skuba config in package.json in a future PR?

PR #1131, opened 2023-03-21
Don't pass explicit list of extensions and let ESLint infer them from the config

mrm007:
As of v7, ESLint is a bit more friendly:

> ESLint will now lint files with extensions other than `.js` if they are explicitly defined in `overrides[].files` - no need to use the `--ext` flag!

So I added a few new files to verify that it still works without providing a list of extensions. You'll notice a few integration tests failing with this message:

```
<random>/a/a/a.mjs
  0:0  error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/a/a/a.mjs` using `parserOptions.project`: /users/rmate/work/skuba/integration/lint/fixable-bd1e312d13bfc2883c3abec50ef3784eb5e55fe52034b08ce2df4892971d4ce6/tsconfig.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file
```

That's because the [ESLint config](https://github.com/seek-oss/eslint-config-skuba/blob/master/index.js#L22) needs to be changed as well. And arguably, this is (more) correct since `typescript-eslint` shouldn't be parsing `.{js,mjs,cjs,jsx}` files.

```diff
diff --git a/index.js b/index.js
index 88db565..f74d054 100644
--- a/index.js
+++ b/index.js
@@ -19,7 +19,7 @@ module.exports = {
       extends: [
         'plugin:@typescript-eslint/recommended-requiring-type-checking',
       ],
-      files: [`*.{${allExtensions.join(',')}}`],
+      files: [`*.{${tsExtensions.join(',')}}`],
       parserOptions: {
         project: './tsconfig.json',
       },

```

If you make the change locally you'll see the tests passing. If you're happy with the changes in this PR I can create a PR in `eslint-config-skuba` with the above diff.

### DIFF
--- a/src/cli/__snapshots__/format.int.test.ts.snap
+++ b/src/cli/__snapshots__/format.int.test.ts.snap
@@ -3,10 +3,10 @@
 exports[`fixable 1`] = `
 "
 ESLint
-Processed 2 files in <random>s.
+Processed 4 files in <random>s.
 
 Prettier
-Processed 6 files in <random>s.
+Processed 8 files in <random>s.
 Formatted 4 files:
 b.md
 c.json
@@ -118,7 +118,7 @@ Processed 2 files in <random>s.
   8:3  error  Unexpected console statement  no-console
 
 <random>/d.js
-  1:0  error  Parsing error: Declaration or statement expected
+  1:0  error  Parsing error: Unexpected token (1:0)
 
 ✖ 2 problems (2 errors, 0 warnings)
 
--- a/integration/base/fixable/a/a/a.mjs
+++ b/integration/base/fixable/a/a/a.mjs
@@ -1,0 +1,7 @@
+// Imports out-of-order
+import path from 'path';
+import fs from 'fs';
+
+export const main = async () => {
+  await fs.promises.access(path.join('.', 'a.mjs'));
+};
--- a/integration/base/fixable/a/a/a.mts
+++ b/integration/base/fixable/a/a/a.mts
@@ -1,0 +1,7 @@
+// Imports out-of-order
+import path from 'path';
+import fs from 'fs';
+
+export const main = async () => {
+  await fs.promises.access(path.join('.', 'a.mts'));
+};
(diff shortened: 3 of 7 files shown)


72636c (approved):
Oh nice, thanks for the cleanup! 🧹

> this is (more) correct since typescript-eslint shouldn't be parsing .{js,mjs,cjs,jsx} files.

FWIW it's _possible_ to use `typescript-eslint` with JavaScript files:

```js
export const main = () => {
  /** @type null */
  const nil = JSON.parse('null');

  // Otherwise:
  // Unsafe return of an `any` typed value.
  // eslint(@typescript-eslint/no-unsafe-return)
  return nil;
};
```

But I can't imagine anyone is getting value out of this, so I agree that we should remove it 👍

mrm007:
@72636c that's a good point that I missed. In that case, I suppose it makes sense to add the files to `"include"` for that project's `tsconfig.json`, right? I tested that and it seemed to work.

72636c:
👍 I'm fine with either approach really. I think it's perfectly reasonable to turn off typescript-eslint on vanilla JS files given we shouldn't have much of that left at SEEK and the remainders are probably config files.

src/cli/__snapshots__/format.int.test.ts.snap
```diff
@@ -118,7 +118,7 @@ Processed 2 files in <random>s.
   8:3  error  Unexpected console statement  no-console
 
 <random>/d.js
-  1:0  error  Parsing error: Declaration or statement expected
+  1:0  error  Parsing error: Unexpected token (1:0)
```
mrm007, line 121: The old error was reported by `typescript-eslint`. This new error is reported by ESLint directly.

samchungy (approved):

PR #1163, opened 2023-05-05
Add an option to preserve assets in the output directory

mrm007:
This is needed for packages which ship Vocab translations, but can be useful for packages which publish CSS, images etc.

### How it works

There's a new config key named `assets`:

```jsonc
// package.json
  "skuba": {
    "assets": [
      "**/*.vocab/*translations.json"
    ],
    // ...
  }
```

This will instruct skuba to copy the files matching the list of globs to the output directory/ies, preserving the directory structure from `src`:
- for `skuba build-package` it will copy them to `lib-commonjs` and `lib-es2015`
- for `skuba build` it will copy them to `lib`

### Context behind this change

This came about when we started the work of [compiling Metropolis packages][compile-metropolis]. As part of that initiative, we've made some changes to Vocab. As of [version 1.2.0][webpack-release] of the Webpack plugin, it can now process compiled code, meaning that it will look for `translations.json` files next to the generated _compiled_ `.vocab/index.ts` file.

That means that consumers a package built with skuba (and SEEK has a few) will start seeing [failing builds] because the package does not ship `translations.json` files. This PR fixes that, but it is not **the** solution. Vocab was not meant to work with compiled code, but this will work for now. We'll need to revisit how Vocab works in the future.

[compile-metropolis]: https://seekchat.slack.com/archives/C014D1WRLJ0/p1682574433645429
[webpack-release]: https://github.com/seek-oss/vocab/releases/tag/%40vocab%2Fwebpack%401.2.0
[failing builds]: https://buildkite.com/seek/metropolis/builds/14847#0187d601-5cb8-44ef-9342-2dfea82c278c/341-394

### Some notes about this PR

There were a few threads on Slack with @askoufis, @72636c and @georgespyropoulos about the best way to do this. We decided to go with the simplest solution for now, which is to add a new `assets` key to `package.json#skuba` which will tell skuba which assets to copy to the output directory/ies. This is similar to how `package.json#files` works.

Currently, these's no documentation about this new feature, but we'll add it soon. What is the best place to document it? The `skuba` config is not documented anywhere at the moment. I thought about adding it in `skuba init` and/or `skuba configure`. @72636c suggested it might be time for a proper `skuba.config.ts`, especially if we want to support some defaults for Vocab.
For now, you can see the tests for how it works.

When using `skuba build` with `"build": "esbuild"` in the `skuba` config, esbuild will not compile `.vocab/index.ts` translation files generated by Vocab. That was a bit weird, so it must be something telling TypeScript/esbuild to ignore those files. I didn't dig deeper and it's unrelated to this PR, but something that would need to be addressed before the next major release [when esbuild becomes the default](https://github.com/seek-oss/skuba/pull/1163/files#diff-468ed41477a6266f0ffba06b439e04998c66e15840fcad715b069651a115ff4fR28). **Edit: fixed in https://github.com/seek-oss/skuba/pull/1164**


askoufis (approved):
LGTM 👍

src/cli/build/assets.ts
```diff
@@ -0,0 +1,79 @@
...
+  getConsumerManifest,
+  getPropFromConsumerManifest,
+} from '../../utils/manifest';
+
+export const copyAssets = async (
+  destinationDir: string,
+  logger: Logger = log,
+) => {
+  const manifest = await getConsumerManifest();
+  const assets = await getPropFromConsumerManifest<string, string[]>('assets');
```
samchungy, line 20: Should we do it in a promise all? Or check one, give up early?

src/cli/build/assets.ts
```diff
@@ -0,0 +1,79 @@
...
+    path.dirname(manifest.path),
+    destinationDir,
+  );
+  const allFiles = await crawlDirectory(resolvedSrcDir);
+  const filesByPattern = buildPatternToFilepathMap(assets, allFiles, {
+    cwd: resolvedSrcDir,
+    dot: true,
+  });
+
+  for (const filenames of Object.values(filesByPattern)) {
```
samchungy, line 37: Would a await promise.all map be better here?

samchungy (approved):
Lgtm otherwise nice one 🙏 @72636c you're probably more across this code

src/cli/build/index.ts
```diff
@@ -43,4 +44,13 @@ export const build = async (args = process.argv.slice(2)) => {
...
   }
+
+  const parsedCommandLine = readTsconfig(args, log);
+
+  if (!parsedCommandLine || process.exitCode) {
+    return;
+  }
+
+  const { options: compilerOptions } = parsedCommandLine;
+  await copyAssets(compilerOptions.outDir ?? 'lib');
```
72636c, line 55: Should we skip this step if there is no `outDir` rather than defaulting to `lib`?

src/cli/build/assets.ts
72636c: As an alternative, it may make sense to flatten `filesByPattern` and deduplicate the result beforehand

src/cli/build/assets.ts
```diff
@@ -0,0 +1,79 @@
...
+  logger: Logger = log,
+) => {
+  const manifest = await getConsumerManifest();
+  const assets = await getPropFromConsumerManifest<string, string[]>('assets');
+
+  if (!manifest || !assets) {
+    return;
+  }
+
+  const resolvedSrcDir = path.resolve(path.dirname(manifest.path), 'src');
```
72636c, line 26: This is interesting as we haven't taken a strong opinion on `src` within the tooling itself, it's just the templated default. Would it make sense for `assets` to be specified relative to the parent directory (alongside `package.json`)?

72636c (approved):
This is great. I'd like to raise the `skuba.config.ts` proposal for review but that can be a separate PR 👍

src/cli/build/assets.ts
mrm007: I've made the change here [`3a6232e` (#1163)](https://github.com/seek-oss/skuba/pull/1163/commits/3a6232ee4a0a18e176f22fd7fc76040e41bfda07). I think a better approach would be to use the directory from `skuba.entryPoint`. What do you think?

PR #1164, opened 2023-05-07
Compile files and directories whose names start with `.`

mrm007:
Found this issue while testing #1163


template/base/tsconfig.build.json
```diff
@@ -1,5 +1,12 @@
 {
   "exclude": ["**/__mocks__/**/*", "**/*.test.ts", "src/testing/**/*"],
   "extends": "./tsconfig.json",
-  "include": ["src/**/*"]
+  "include": [
+    "src/**/*",
+    // Must be explicit about including dotfiles otherwise TypeScript ignores them
+    // https://github.com/microsoft/TypeScript/blob/v5.0.4/src/compiler/utilities.ts#L8828-L8830
```
72636c, line 7: Thanks for finding this @mrm007. I recall you first mentioned that this affected our `esbuild` implementation; does it affect `tsc` too?

If it's `esbuild` only then I'm hoping we can add some adapter logic to deal with this under the hood.

mrm007: Strangely, it's only for esbuild. Running `tsc` via the command line finds and compiles all files, as expected.

The filter is applies only when resolving the config programmatically.

72636c: Okay wow, this is a _great_ find. It comes down to this and is actually a deeper bug with the `esbuild` implementation:

https://github.com/seek-oss/skuba/blob/3196eb130ef846e902bb8dd429e3b0d2d9418c30/src/cli/build/esbuild.ts#L68-L82

The TypeScript API does not include dot paths when deriving `fileNames` and this holds true for `tsc` internally. However, `tsc` performs module resolution and will subsequently include files like `.vocab/index.ts` only if they have been imported by another module in `fileNames`. On the other hand, our `esbuild` adapter disables bundling and ends up classifying files like `.vocab/index.ts` as "external" dependencies that should not be included.

There are a few options here:

1. Ask consumers to manually modify their `tsconfig.json#include` as per this PR. This is simple but feels like a rather arcane fix.
1. Use a custom method to derive `fileNames` in the `esbuild` adapter that includes dot paths by default. This would let us play nicely with Vocab but won't promise to match the TypeScript implementation.
2. We can read the `metafile` out of `esbuild` and identify imports that should be included in a second pass.

mrm007: What if we do the globbing ourselves and include all the dotfiles? Basically, applying the same technique mentioned here https://github.com/seek-oss/skuba/pull/1163#discussion_r1190497109:
- take the root directory for `entryPoint` i.e. `src` for `src/index.ts`
- find all `*.{ts,tsx,cts,mts}` files
- compile them with esbuild

Something like this [`2c96e5b` (#1164)](https://github.com/seek-oss/skuba/pull/1164/commits/2c96e5b6b65950dcc2b7d1c830affcc753299acb)

PR #1168, opened 2023-05-09
Propose `skuba.config.ts`

### LINKED ISSUE #1167: Add skuba.config.ts
This is great. I'd like to raise the `skuba.config.ts` proposal for review but that can be a separate PR 👍

_Originally posted by @72636c in https://github.com/seek-oss/skuba/pull/1163#pullrequestreview-1419540213_

72636c:
Steals from sku and ESLint's recent direction. Follows on from #1163 and resolves #1167.

### DIFF
--- a/src/config/load.ts
+++ b/src/config/load.ts
@@ -1,0 +1,32 @@
+import path from 'path';
+
+import { register } from 'esbuild-register/dist/node';
+
+import { type SkubaConfig, skubaConfigSchema } from './types';
+
+import { getDestinationManifest } from 'cli/configure/analysis/package';
+import { log } from 'utils/logging';
+
+const CONFIG_FILENAME = 'skuba.config.ts';
+
+export const loadSkubaConfig = async (cwd?: string): Promise<SkubaConfig> => {
+  // FIXME: remove underlying `process.exit`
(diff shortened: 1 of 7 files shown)


src/config/load.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const CONFIG_FILENAME = 'skuba.config.ts';
+
+export const loadSkubaConfig = async (cwd?: string): Promise<SkubaConfig> => {
+  // FIXME: remove underlying `process.exit`
+  const manifest = await getDestinationManifest({ cwd });
+
+  const configPath = path.join(manifest.path, '..', CONFIG_FILENAME);
+
+  const esbuildRegistration = register({ target: 'node16' });
```
mrm007, line 18: It might be worth having a hot reloading ability when the config changes, e.g. restart the server started with `skuba start` or anything like that.

For reference, we do this in Crackle: 
https://github.com/seek-oss/crackle/blob/49b48ee405477cccdc1d6e7bbbb4c26b921c7760/packages/cli/src/index.ts#L33-L44

tadhglewis:
Any update on whether we're going ahead with these changes? Let's snooze this for now otherwise

PR #1173, opened 2023-05-17
Update GitHub Annotation type

samchungy:
The Annotation type for the patch is the same as the create. Should fix our issues while we wait for them: https://github.com/octokit/openapi-types.ts/issues/313


samchungy:
Oops looks like still broken nvm

PR #1174, opened 2023-05-17
Fix @octokit checkRun annotation type

samchungy:
https://github.com/octokit/openapi-types.ts/issues/313

Use an alternative type to fix the Octokit types. The `annotation` type is the same on the patch as the get.

More and more consumers are running into and they are taking a while to fix the issue:

https://seekchat.slack.com/archives/C03UM9GBGET/p1684126680976739
https://seekchat.slack.com/archives/C03UM9GBGET/p1682467361331139


72636c (approved):
🙇

PR #1175, opened 2023-05-17
Enable `noUncheckedIndexedAccess`

samchungy:
Release seek-tsconfig 2.0


src/cli/test/reporters/github/index.test.ts
```diff
@@ -120,14 +120,16 @@ const successResults: AggregatedResult = {
...
+  testResults: failResults.testResults[0]
+    ? [
+        {
+          ...failResults.testResults[0],
+          numFailingTests: 0,
+          numPassingTests: 1,
+          testResults: [],
+        },
+      ]
+    : [],
```
samchungy, line 132: This isn't quite equivalent (will revisit)

.changeset/healthy-mirrors-look.md
```diff
@@ -0,0 +1,35 @@
+---
+'skuba': minor
```
samchungy, line 2: minor or major? 🤔

.changeset/healthy-mirrors-look.md
72636c, line 1: Oof, I'm kinda hot and cold on this change now that the tsconfig-seek upgrade has propagated to a couple of our repos and I'm seeing this PR as well 😅. It seems like there are many common scenarios that are not inferred. For example:

```typescript
const result = /^([a-z])/.match(input);

if (result) {
  // !!!
  result[1];
}

// !!!
''.split('#')[0];

const xs: Array<{key: string, value: string}>;

if (xs.length === 1) {
  // !!!
  const x = xs[0];

  const record = Object.fromEntries(xs.map(x) => [x.key, x.value]);

  // !!!
  record[x.key];
}
```

I don't blame TypeScript here but I worry about the DX on generally well-behaved codebases once we enforce this option. We need to make sure that we can somehow deliver this _without_ normalising usage of easy escape hatches like [`!`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator); what we don't want is for this change to end up proliferating unsafe practices, and beyond just indexed access.

template/lambda-sqs-worker/src/app.ts
```diff
@@ -36,11 +36,17 @@ export const handler = createHandler<SQSEvent>(async (event) => {
 
   const record = event.Records[0];
 
+  const body = record?.body;
+
+  if (!body) {
+    throw new Error(`Record: ${JSON.stringify(record)} is malformed`);
+  }
```
72636c, line 43: I'm a bit cautious of logging the whole `record`. Could we do something like:

```suggestion
  if (!record) {{
    throw new Error('Malformed SQS event with no records');
  }
  
  // Or just refer to `record.body` below
  const { body } = record;
```

src/cli/init/getConfig.ts
```diff
@@ -156,6 +156,10 @@ const baseToTemplateData = async ({
 
   const port = String(await getRandomPort());
 
+  if (!orgName) {
+    throw new Error('Could not determine an ownerName');
```
72636c, line 160:
```suggestion
    throw new Error(`Invalid format for owner name: ${ownerName}`);
```

src/cli/configure/getEntryPoint.ts
```diff
@@ -42,6 +42,10 @@ export const getEntryPoint = ({
       // Support exported function targeting, e.g. `src/module.ts#callMeMaybe`
       const [modulePath] = value.split('#', 2);
 
+      if (!modulePath) {
+        return `${chalk.bold(value)} is an invalid module path`;
+      }
```
72636c, line 47: Is this even possible? The only way I've been able to arrive at an empty array is by doing `''.split('')`.

src/api/git/reset.int.test.ts
```diff
@@ -44,7 +44,7 @@ describe('soft', () => {
 
     const directory = await fs.promises.readdir(dir);
 
-    expect(commits[0].oid).toEqual(initialCommit);
+    expect(commits[0]?.oid).toEqual(initialCommit);
```
72636c, line 47: It's unfortunate that this applies to tests too; ideally we'd have an ESLint rule so we could disable it on test files.

It's a bit of a wash in this example, but I'd be open to use of `!` in tests. A runtime error is harmless here and would equally indicate that the test failed.

scripts/package.ts
```diff
@@ -171,8 +171,12 @@ const compileChangesByTemplate = (changelog: string) => {
...
         // Note that the changeset entry applies to the template if it existed
         // as of this version and the scope is a match.
-        if (semver.gt(version, added) && templateMatcher.test(templateName)) {
-          changesByTemplate[templateName].push(
+        if (
+          semver.gt(version, added) &&
+          templateMatcher.test(templateName) &&
+          changesByTemplate[templateName]
+        ) {
+          changesByTemplate[templateName]!.push(
```
72636c, line 179:
```suggestion
          changesByTemplate[templateName].push(
```

72636c (approved):

72636c (approved):

PR #1181, opened 2023-05-26
Disable noUnusedLocals and noUnusedParameters in tsconfig

eddeee888:
Make `noUnusedLocals` and `noUnusedParameters` false.

These two values are set as `true` in SEEK's tsconfig.

Skuba is already using Seek's ESLint config which has a [rule](https://eslint.org/docs/latest/rules/no-unused-vars) which works for both function and types.

We do not need both tools to do the same thing. ESLint has better support for ignoring files  if needed.

### DIFF
--- a/config/tsconfig.json
+++ b/config/tsconfig.json
@@ -3,7 +3,9 @@
     "incremental": true,
     "isolatedModules": true,
     "moduleResolution": "node",
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "noUnusedLocals": false,
+    "noUnusedParameters": false
   },
   "extends": "tsconfig-seek"
 }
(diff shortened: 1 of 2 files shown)


config/tsconfig.json
72636c, line 1: Thanks for the PR @eddeee888! Would it make sense to update https://github.com/seek-oss/tsconfig-seek/blob/0aa686106e7a2a3d9c7e420b88d208932126d58d/tsconfig.json#L34-L35 instead?

samchungy: https://seekchat.slack.com/archives/C03UM9GBGET/p1685060237597449?thread_ts=1685057121.273509&cid=C03UM9GBGET

samchungy: Happy to go either way though

72636c (approved):
🙇

PR #1223, opened 2023-07-19
Remove 'warm' build steps from the template pipelines

gitlon:
Remove 'warm' build steps from the template pipelines, because they don't have any effect in the new `vCurrent` build agent queues.

These queues don't run steps on the same EC2 instance so there is warming, pre-launching or other time-saved, only an uneccessary cost to run these.

This change does not change Skuba functionality, nor the build-compile of Skuba iteself.


template/express-rest-api/.buildkite/pipeline.yml
```diff
@@ -30,21 +30,12 @@ configs:
           permit_on_passed: true
 
 steps:
-  - label: 🧖‍♀️ Warm Prod
-    command: ':'
-    key: warm-prod
-    plugins:
-      - *aws-sm
-      - *private-npm
-      - *docker-ecr-cache
```
72636c, line 39: The intent here was to bake the Docker build image and uploading it to ECR once, rather than doing it twice on `Test & Lint` and `Build & Package`. Thoughts?

samchungy: Yeah in my team's repo we call the step `build cache`

gitlon: Yes of course that is still valid.

How about serialising two steps, so it gets done only once - putting the shortest step first (which I think will be `Build & Package`)? Or am I just wrong, and the warm step should remain.

samchungy: If both run at the same time there would be a chance we would get a cache miss right?

gitlon: By "serialise" I meant adding a dependency so that one step always runs after the other, so no chance of a cache miss

72636c: I think the concern is that a serial approach could work out to be slower overall:

```
Cache 5m
Build 2m | Test 5m = 5m
10m

Build 5m + 2m = 7m
Test 5m
12m
```

It would be great if we had data to substantiate how much we stand to gain/lose in practice. Ideally this would be roughly break even and we can go with the simpler serial approach.

samchungy: Not to mention it might fail while quicker while running both lint and build at the same time.

gitlon: I am investigating / gathering stats.

samchungy: Right, as of the new build queue it looks like each step has to do the pull image step, so the question is whether that pull image step is quicker than running the commands on a single step

AaronMoat: It would be nice if we could somehow lock the ecr cache plugin to only run once per build at a time, then the first one that runs will build + the others will pull, without the wait step in other scenarios. I am unclear if that is remotely possible though.

samchungy: This definitely helps in UI image, given the UI images are typically alot bigger.

It takes on average 1 min 30 seconds to pull the image on one of our UI repos so moving all the commands to the single step has sped it up a bit

gitlon:
Per the discussion, this isn't the right change. Sam made a better change by updating the docker-ecr-cache plugin and using the new `skip-pull-from-cache` here.

PR #1224, opened 2023-07-21
Update distroless nodejs image

samchungy:
The current image has these vulnerabilities flagged in ECR but the new `-debian11` image is free from them.

image

For whatever reason, this image isn't using the latest deps/libs? I haven't quite looked at why but the documentation now recommends `gcr.io/distroless/nodejs18-debian11` by default

https://github.com/GoogleContainerTools/distroless/blob/main/nodejs/README.md

### DIFF
--- a/src/cli/configure/patchDockerfile.ts
+++ b/src/cli/configure/patchDockerfile.ts
@@ -1,0 +1,38 @@
+import { inspect } from 'util';
+
+import fs from 'fs-extra';
+
+import { log } from '../../utils/logging';
+
+import { createDestinationFileReader } from './analysis/project';
+
+const DOCKERFILE_FILENAME = 'Dockerfile';
+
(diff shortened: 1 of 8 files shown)


src/cli/configure/patchDockerfile.ts
samchungy, line 1: Should we autopatch this or just tell consumers to? 🤷

72636c: We probably need to agree on a more systematic approach for introducing and later retiring custom patches like this, but makes a lot of sense here 👍

72636c (approved):

src/cli/configure/patchDockerfile.ts
samchungy: Re this, I'm thinking maybe with every major version we retire custom patches. Somewhere in the readme we should add a section on upgrading and suggest that consumers go up major version's one at a time perhaps?

PR #1241, opened 2023-08-15
Fix Patch Renovate Config for `skuba init`

samchungy:
Noticed this when I was running `skuba init`

Not urgent as it can be resolved with a `yarn format` after it's been initialised.

### DIFF
--- a/.changeset/tiny-chairs-deliver.md
+++ b/.changeset/tiny-chairs-deliver.md
@@ -1,0 +1,9 @@
+---
+'skuba': minor
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/tiny-chairs-deliver.md
```diff
@@ -0,0 +1,9 @@
+---
+'skuba': minor
```
72636c, line 2: Ta! Should we `patch` this?

72636c (approved):

PR #1269, opened 2023-09-25
Fix autofixes run on a nested directory

72636c:
Context for SEEKers: https://seekchat.slack.com/archives/C03UM9GBGET/p1695369614026749


src/api/git/commitAllChanges.int.test.ts
```diff
@@ -78,6 +87,42 @@ it('should stage and commit removed files', async () => {
...
+
+  const statuses = await Promise.all([
+    git.status({ fs, dir, filepath: newFileName }),
+    git.status({ fs, dir, filepath: newFileName2 }),
+    git.status({ fs, dir, filepath: path.join(nestedDir, newFileName) }),
+    git.status({ fs, dir, filepath: path.join(nestedDir, newFileName2) }),
+  ]);
+
+  // The file outside of our `nestedDir` remains uncommitted.
+  expect(statuses).toStrictEqual(['absent', '*added', 'unmodified', 'absent']);
```
72636c, line 123: Does this check out as reasonable design? If you run `skuba lint` in `/packages/package`, I assume it would be surprising for a file in another directory to get committed as part of autofixes.

The overarching counterpoint is that running `skuba lint` on each subdirectory separately is actually problematic for autofixing in general. What we really need is a single autofix commit to be produced per repository, not per subdirectory.

docs/development-api/github.md
```diff
@@ -128,7 +128,7 @@ and maps them to GitHub GraphQL [FileChanges].
 ```typescript
 import { GitHub } from 'skuba';
 
-const fileChanges = await GitHub.readFileChanges([
+const fileChanges = await GitHub.readFileChanges(dir, [
```
72636c, line 131: `ChangedFiles` are relative to `dir` so we really need both to correctly evaluate `FileChanges`

src/api/git/commitAllChanges.int.test.ts
72636c: I think this may be the best incremental fix we can put in place for now, but it ties in to a broader investigation into how we better support monorepos:

- If a monorepo contains only skuba projects: it may be best to run `skuba lint` at the repository root
- If a monorepo contains a mix of projects including 1 skuba one: this PR works well
- If a monorepo contains a mix of projects including >1 skuba one: this PR helps but you'd still be running X `lint`s which would try to push X separate autofix commits

samchungy: Makes sense. Multiple commits in it's current state also won't work either right now unless we calculate a dynamic message

samchungy (approved):

Issue #603, opened 2021-10-07
Audit zero-config mode


72636c:
We recommend a global install, so it would nice if you could bring skuba into an existing TypeScript project/directory you have and start using its commands without having to run a heavyweight `skuba configure` migration.

This may require synthesising some of the usual config and ignore files we depend on upfront, either to disk or within our process.

PR #1274, opened 2023-09-30
Skuba Patches

samchungy:
Brought up in a previous PR.

We need a plan to manage when we run patches as at the moment we are always running them when realistically they only ever need to be run once.

By updating the skuba manifes version we can essentially see what version they were previously on versus now.

The version number in the `patches` folder is the `from` version so we don't need to know what the new version number will be.


src/cli/__snapshots__/format.int.test.ts.snap
```diff
@@ -7,11 +7,10 @@ Processed 4 files in <random>s.
 
 Prettier
 Processed 8 files in <random>s.
-Formatted 4 files:
+Formatted 3 files:
```
samchungy, line 10: Gets formatted by the upgrade so it doesn't appear here

package.json
```diff
@@ -165,6 +165,6 @@
     "entryPoint": "src/index.ts",
     "template": null,
     "type": "package",
-    "version": "4.0.0"
+    "version": "7.3.1"
```
samchungy, line 168: This may or may not be bumped up in the version packages PR

src/cli/lint/autofix.ts
```diff
@@ -12,10 +12,10 @@ import { createLogger, log } from '../../utils/logging';
 import { throwOnTimeout } from '../../utils/wait';
 import { runESLint } from '../adapter/eslint';
 import { runPrettier } from '../adapter/prettier';
-import { JEST_SETUP_FILES } from '../configure/addEmptyExports';
 import { RENOVATE_CONFIG_FILENAMES } from '../configure/modules/renovate';
-import { SERVER_LISTENER_FILENAME } from '../configure/patchServerListener';
 import { REFRESHABLE_IGNORE_FILES } from '../configure/refreshIgnoreFiles';
+import { JEST_SETUP_FILES } from '../configure/upgrade/patches/7.3.1/addEmptyExports';
```
AaronMoat, line 17: Is it right for this to be imported from a patch file? Similar comments apply elsewhere

src/cli/configure/upgrade/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    getConsumerManifest(),
+  ]);
+
+  const manifestVersion = (
+    manifest?.packageJson.skuba as SkubaPackageJson | undefined
+  )?.version;
+
+  if (!manifest || !manifestVersion) {
+    throw new Error(
+      'Could not find a skuba manifest, please run `skuba configure`',
```
AaronMoat, line 31: Wonder how many people would be missing it but are genuinely using skuba? Could it gracefully configure automatically?

src/cli/configure/upgrade/index.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    jest.mocked(getSkubaVersion).mockResolvedValue('2.0.0');
+
+    // readdir has overloads and the mocked version doesn't match the string version
+    jest
+      .mocked(readdir)
+      .mockResolvedValue(['0.9.0', '1.0.0', '2.0.0'] as unknown as Dirent[]);
+
+    await expect(upgradeSkuba()).resolves.toBeUndefined();
+    expect(mockUpgrade.upgrade).toHaveBeenCalledTimes(2);
+    //     expect(writeFile).toHaveBeenCalledWith(
```
AaronMoat, line 94: wip?

AaronMoat (commented):
Also, how does lint calling upgradeSkuba work - would it still write?

src/cli/lint/autofix.ts
samchungy: Yeah I dunno, happy to get some 2 cents. I'm thinking about possibly making all import from here?

src/cli/init/index.ts
```diff
@@ -10,7 +10,7 @@ import {
   BASE_TEMPLATE_DIR,
   ensureTemplateConfigDeletion,
 } from '../../utils/template';
-import { tryPatchRenovateConfig } from '../configure/patchRenovateConfig';
+import { tryPatchRenovateConfig } from '../configure/upgrade/patches/7.3.1/patchRenovateConfig';
```
AaronMoat, line 13: Do we definitely want this in init?

src/cli/lint/autofix.ts
AaronMoat: If from here you mean autofix.ts I think that's fine

samchungy:
> Also, how does lint calling upgradeSkuba work - would it still write?

Technically yes, it already does this so it wouldn't be a functional change 😅

src/cli/configure/upgrade/index.ts
samchungy: The problem with that is you have to pick a template I believe 🤔

samchungy:
Late night thoughts: maybe I should add some logs to imply that we're patching their files. Eg. Updating skuba to X, Y patch applied.

AaronMoat (approved):
Logging sounds good

src/cli/format.ts
```diff
@@ -5,20 +5,11 @@ import { createLogger, log } from '../utils/logging';
 
 import { runESLint } from './adapter/eslint';
 import { runPrettier } from './adapter/prettier';
-import { tryAddEmptyExports } from './configure/addEmptyExports';
-import { tryPatchDockerfile } from './configure/patchDockerfile';
-import { tryPatchRenovateConfig } from './configure/patchRenovateConfig';
-import { tryPatchServerListener } from './configure/patchServerListener';
 import { tryRefreshIgnoreFiles } from './configure/refreshIgnoreFiles';
```
72636c, line 8: Is it worth modelling this as an "evergreen upgrade", so the commands only have to worry about invoking `upgradeSkuba` and we could introduce other upgrades of its nature in future?

src/cli/configure/upgrade/patches/7.3.1/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import { tryAddEmptyExports } from './addEmptyExports';
+import { tryPatchDockerfile } from './patchDockerfile';
+import { tryPatchServerListener } from './patchServerListener';
+
+export const upgrade = async () => {
+  await Promise.all([
+    tryAddEmptyExports(),
+    tryPatchRenovateConfig(),
+    tryPatchDockerfile(),
+    tryPatchServerListener(),
```
72636c, line 12: This is a bit scope creepy, but thoughts on establishing a convention that patches need to describe what they are doing, e.g. which files are affected, how, and why? This could involve either or both of:

- Console logging
- Commit messaging that is propagated to the autofix

Happy to raise this (creating guidelines for contributing patches) as a separate issue if we agree it's a good idea, doesn't have to be solved in this initial PR 👍

src/cli/configure/upgrade/index.test.ts
```diff
@@ -0,0 +1,138 @@
+import type { Dirent } from 'fs';
+import { readdir, writeFile } from 'fs/promises';
```
72636c, line 2: Should we stick to `fs-extra`?

I initially thought of dropping it from skuba, but its use of `graceful-fs` under the hood may be a worthwhile improvement over the standard module: https://github.com/isaacs/node-graceful-fs#graceful-fs

src/cli/configure/upgrade/index.test.ts
```diff
@@ -0,0 +1,138 @@
...
+        name: 'some-api',
+        readme: '',
+        version: '1.0.0',
+      } as NormalizedPackageJson,
+      path: '/package.json',
+    });
+
+    await expect(upgradeSkuba()).rejects.toThrow(
+      'Could not find a skuba manifest, please run `skuba configure`',
+    );
```
72636c, line 34: Is this a desirable breaking change? If so, should we warn and skip patching until the next major version?

One consideration is that this may move us further away from #603. I like zero-config support as an aspiration, but appreciate that we'd have to evaluate the tradeoffs to get us there.

src/cli/configure/upgrade/index.ts
```diff
@@ -0,0 +1,82 @@
...
+  if (gte(manifestVersion, currentVersion)) {
+    return;
+  }
+
+  log.newline();
+  log.plain('Updating skuba...');
+
+  const patches = await getPatches(manifestVersion);
+
+  // Run these in series in case a previous patch relies on another patch
```
72636c, line 66:
```suggestion
  // Run these in series in case a subsequent patch relies on a previous patch
```

src/cli/configure/upgrade/index.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  // The patches are sorted by the version they were added from.
+  // Only return patches that are newer or equal to the current version.
+  return sort(patches.filter((filename) => gte(filename, manifestVersion)));
+};
+
+const fileExtensions = ['js', 'ts'];
+
+export type Patch = {
+  upgrade: () => Promise<void>;
```
72636c, line 23: Just thinking, would a patch ever want/need to know any metadata about the current project? For example, could we introduce a patch that should only be applied to projects templated from `lambda-sqs-worker-cdk`?

src/cli/configure/upgrade/index.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+// Hack to allow tests to resolve the patches
```
72636c, line 26: Nitpick: should we call out Jest config here specifically:

```suggestion
// Hack to allow our Jest environment/transform to resolve the patches
```

72636c (approved):
Brill!

src/cli/build/index.ts
```diff
@@ -3,15 +3,12 @@ import chalk from 'chalk';
...
 import { log } from '../../utils/logging';
 import { getStringPropFromConsumerManifest } from '../../utils/manifest';
-import { tryAddEmptyExports } from '../configure/addEmptyExports';
 
 import { copyAssets } from './assets';
 import { esbuild } from './esbuild';
 import { readTsconfig, tsc } from './tsc';
 
 export const build = async (args = process.argv.slice(2)) => {
-  await tryAddEmptyExports();
```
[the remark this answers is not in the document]
samchungy: Sounds good to me

src/cli/configure/upgrade/index.test.ts
samchungy: I suppose we only require a simple `{ skuba: { version: string } }` field. Maybe we can just auto configure the bare minimum?

src/cli/configure/upgrade/index.ts
samchungy: Eh we can add it down the line if we decide to :P

72636c: Yep not for this PR, just thinking about the thinking :)

src/cli/format.ts
samchungy: Hmmm I'll leave it for now 🤔

PR #1278, opened 2023-10-01
template: Adjust `.npmrc` handling

72636c:
https://docs.npmjs.com/cli/v10/configuring-npm/npmrc#files


template/greeter/Dockerfile
```diff
@@ -7,5 +7,5 @@ WORKDIR /workdir
 COPY package.json yarn.lock ./
```
72636c, line 7: We would include `.npmrc` in this `COPY` once we switch the templates over to pnpm

samchungy:
Given .npmrc can now affect the node_modules, should we also set `.npmrc` as one of the cache keys?

72636c:
@samchungy Reckon we should land this here, or when we switch the templates over to pnpm?

Probably just comes down to whether this log bothers you in the meantime 😅

Screenshot 2023-10-06 at 12 14 31 pm

samchungy (approved):

PR #1282, opened 2023-10-03
fix supertest app.test.ts example

lukeymyuan:
### Why

The below unit tests do not fail when endpoint returns 404
We either use `expect` function or throw when check fails  https://github.com/ladjs/supertest#expectfunctionres-

### What
Modify unit test
Add usage message when argument is not received

### DIFF
--- a/template/koa-rest-api/src/app.test.ts
+++ b/template/koa-rest-api/src/app.test.ts
@@ -17,10 +17,18 @@ describe('app', () => {
       }
     }));
 
+  /**
+   * Test assertion with logic
+   */
   it('has a reachable nested route', () =>
     agent.get('/jobs').expect(({ status }) => {
-      if (status === 404) {
-        throw new Error('Unreachable nested route');
+      switch (status) {
+        case 200:
+          return;
+        case 404:
+          throw new Error('Unreachable jobs route');
+        default:
+          throw new Error(`Unexpected status: ${status}`);
       }
     }));
 
--- a/template/koa-rest-api/src/app.test.ts
+++ b/template/koa-rest-api/src/app.test.ts
@@ -10,12 +10,10 @@ describe('app', () => {
 
   it('has a happy health check', () => agent.get('/health').expect(200, ''));
 
-  it('has a reachable smoke test', () =>
-    agent.get('/smoke').expect(({ status }) => {
-      if (status === 404) {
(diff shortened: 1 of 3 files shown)


template/koa-rest-api/src/app.test.ts
```diff
@@ -11,10 +11,26 @@ describe('app', () => {
   it('has a happy health check', () => agent.get('/health').expect(200, ''));
 
   it('has a reachable smoke test', () =>
-    agent.get('/smoke').expect(({ status }) => status !== 404));
+    agent.get('/smoke').expect(({ status }) => {
+      if (status === 404) {
+        throw new Error('Unreachable smoke test');
+      }
+    }));
```
72636c, line 18: Thanks for the fixes Luke!

Thoughts on whether this may be cleaner:

```suggestion
  it('has a reachable smoke test', async () => {
    const response = await agent.get('/smoke');

    expect(response.status).not.toBe(404);
  });
```

template/koa-rest-api/src/app.test.ts
```diff
@@ -10,11 +10,25 @@ describe('app', () => {
...
+    const response = await agent.get('/smoke');
+    expect(response.status).not.toBe(404);
+  });
 
+  /**
+   * Test assertion with logic
+   */
   it('has a reachable nested route', () =>
-    agent.get('/jobs').expect(({ status }) => status !== 404));
+    agent.get('/jobs').expect(({ status }) => {
```
72636c, line 22: Can we do the same here? I'd prefer to avoid this `expect(fn)` form entirely given the mistake we previously made 🥲

template/koa-rest-api/src/app.test.ts
```diff
@@ -10,11 +10,25 @@ describe('app', () => {
...
-  it('has a reachable smoke test', () =>
-    agent.get('/smoke').expect(({ status }) => status !== 404));
+  it('has a reachable smoke test', async () => {
+    const response = await agent.get('/smoke');
+    expect(response.status).not.toBe(404);
+  });
 
+  /**
+   * Test assertion with logic
+   */
```
72636c, line 20: I'm not sure I understand the value of this comment. Can we remove it?

72636c (approved):
🙇 thanks a bunch!

Issue #1285, opened 2023-10-09
pnpm migration


samchungy:
Thread to keep track of all pnpm related threads

Lint Hanging: https://github.com/seek-oss/skuba/issues/836
Configure uses yarn: https://github.com/seek-oss/skuba/issues/1246
Manage .npmrc: https://github.com/seek-oss/skuba/issues/1286
Serverless support: https://github.com/seek-oss/skuba/issues/1287

72636c:
Closing this one as the remaining work is pinned and tracked under #1286

PR #1289, opened 2023-10-10
Support pnpm

72636c:
Related: #1285

To do:

- [x] De-yarn Buildkite pipelines
- [x] De-yarn Dockerfiles
- [x] De-yarn `skuba configure` command
- [x] De-yarn CLI logging (we have e.g. "Run yarn skuba configure" in places)


72636c:
Was not expecting a ✅ today, must be something wrong with CI!

template/oss-npm-package/_package.json
```diff
@@ -38,6 +38,7 @@
     "commitizen": "^4.2.4",
     "skuba": "*"
   },
+  "packageManager": "pnpm@8.9.0",
```
72636c, line 41: Worth discussing if we want to pin a particular pnpm version, and if so, whether we should express it in this way.

This is supported by `pnpm/action-setup` in GitHub Actions, but I don't think it'll work with [Corepack](https://pnpm.io/installation#using-corepack) in our Dockerfiles elsewhere.

template/greeter/Dockerfile
```diff
@@ -2,10 +2,14 @@
...
+
 WORKDIR /workdir
 
 COPY package.json yarn.lock ./
 
-RUN \
-  --mount=type=secret,id=npm,dst=/root/.npmrc \
-  yarn install --frozen-lockfile --ignore-optional --non-interactive
+RUN --mount=type=secret,id=npm,dst=/root/.npmrc pnpm fetch
+RUN pnpm install --offline --recursive
```
72636c, line 15: This is largely copied from some of my earlier experimentation; we should see if there are better implementations out there that we should copy instead.

template/lambda-sqs-worker-cdk/package.json
```diff
@@ -6,7 +6,7 @@
     "deploy": "cdk deploy appStack --require-approval never --context stage=${ENVIRONMENT}",
     "format": "skuba format",
     "lint": "skuba lint",
-    "package": "yarn install --ignore-optional --ignore-scripts --modules-folder ./lib/node_modules --non-interactive --offline --production",
+    "package": "pnpm install --ignore-scripts --modules-folder ./lib/node_modules --no-optional --offline --production",
```
72636c, line 9: This package script seems dicey; we'll have to manually validate that we can implement an equivalent with pnpm

template/private-npm-package/README.md
```diff
@@ -82,7 +82,7 @@ If you use the `Squash and merge` option on pull requests, take extra care to fo
 ### Releasing latest
 
 Commits to the `master` branch will be released with the `latest` tag,
-which is the default used when running `npm install` or `yarn install`.
+which is the default used when running `npm install` or `pnpm install`.
```
72636c, line 85: Looks like Gutenberg supports pnpm, but we should try initialising a new package once we have a stable snapshot release to make sure that it works end to end.

template/greeter/Dockerfile
```diff
@@ -2,10 +2,13 @@
 
 FROM --platform=${BUILDPLATFORM:-<%- platformName %>} node:18-alpine AS dev-deps
 
+RUN corepack enable
+RUN corepack prepare pnpm@latest --activate
```
samchungy, line 6: `RUN corepack enable pnpm` seems to do the job of the above two

72636c: Is there some gotcha around the pnpm version? https://pnpm.io/installation#using-corepack

samchungy: Oooo unsure, I'll experiment

samchungy: image

Appears to install latest version

samchungy: image

Appears to respect the `packageManager` from package.json

template/greeter/Dockerfile
```diff
@@ -2,10 +2,12 @@
...
+
 WORKDIR /workdir
 
 COPY package.json yarn.lock ./
 
-RUN \
-  --mount=type=secret,id=npm,dst=/root/.npmrc \
-  yarn install --frozen-lockfile --ignore-optional --non-interactive
+RUN --mount=type=secret,id=npm,dst=/root/.npmrc pnpm fetch
+RUN pnpm install --offline --recursive
```
72636c, line 13: @samchungy Did you end up running `pnpm install` in your Dockerfile?

Trying to figure out the best way forward, especially where we [build a second Dockerfile for Gantry interop](https://github.com/seek-oss/skuba/blob/master/template/koa-rest-api/Dockerfile) 🥲

samchungy: Yeah I opted for simplicity:

https://github.com/SEEK-Jobs/backoffice-budgets-api/pull/224/files#diff-02e5f267970af3f4a889c28bba25f5a39b956dcd12a541a62340bbd06e0cafecL11

template/base/_.npmrc
```diff
@@ -0,0 +1,6 @@
+# managed by skuba
```
72636c, line 1: Review that this doesn't mess with existing `.npmrc`s that should be ignored

src/utils/packageManager.ts
```diff
@@ -0,0 +1,64 @@
...
+    const raw = await detect({ cwd, includeGlobalBun: false });
+
+    packageManager = packageManagerSchema.parse(raw);
+  } catch (err) {
+    log.warn(
+      `Failed to detect package manager; defaulting to ${log.bold(
+        DEFAULT_PACKAGE_MANAGER,
+      )}.`,
+    );
+    log.subtle(inspect(err));
```
72636c, line 54: Adjust this; its too verbose, particularly for `ZodError`s.

template/express-rest-api/Dockerfile
```diff
@@ -5,15 +5,16 @@ ARG BASE_TAG
 
 FROM ${BASE_IMAGE}:${BASE_TAG} AS deps
 
-RUN yarn install --ignore-optional --ignore-scripts --non-interactive --offline --production
+RUN pnpm install --offline --production --recursive
```
samchungy, line 8: I don't think --recursive is needed with pnpm install from when I was testing

template/express-rest-api/Dockerfile.dev-deps
```diff
@@ -2,10 +2,12 @@
...
+RUN corepack enable pnpm
+RUN pnpm config set store-dir /root/.pnpm-store
 
-COPY package.json yarn.lock ./
+WORKDIR /workdir
 
-RUN \
-  --mount=type=secret,id=npm,dst=/root/.npmrc \
-  yarn install --frozen-lockfile --ignore-optional --non-interactive
+RUN --mount=type=bind,source=package.json,target=package.json \
```
samchungy, line 10: We can skip binding on package.json if we're using fetch yeah?

72636c: Ah, and I need to include `.npmrc` instead 👍

samchungy (approved):

template/express-rest-api/.buildkite/pipeline.yml
```diff
@@ -11,8 +11,9 @@ configs:
     - &docker-ecr-cache
       seek-oss/docker-ecr-cache#v2.1.0: &docker-ecr-cache-defaults
         cache-on:
+          - .npmrc
```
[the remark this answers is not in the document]
72636c: Just playing this back for my benefit: we can exclude `package.json` from the cache key because a stale lockfile will fail on the subsequent `pnpm install --offline`?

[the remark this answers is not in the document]
samchungy: Yeah. So our dev deps now only contains a .pnpm-store, it should remain valid if we change lets say the `description` field in our package.json.

template/base/_.npmrc
72636c: https://github.com/seek-oss/skuba/pull/1329#issuecomment-1859528370

template/private-npm-package/README.md
72636c: https://github.com/seek-oss/skuba/pull/1329#issuecomment-1859528370

PR #1294, opened 2023-10-11
Use Jest projects

samchungy:
Utilise and spruik Jest projects because it's pretty

image

Unfortunately running integration tests only in band via projects is not supported 😢 

https://github.com/jestjs/jest/issues/10936


jest.config.ts
```diff
@@ -1,12 +1,34 @@
...
+export default {
+  ...defaultConfig,
+  projects: [
+    {
+      transform,
+      moduleNameMapper,
+      setupFiles,
+      displayName: 'unit',
+      testPathIgnorePatterns: ['<rootDir>/template/', '\\.int\\.test\\.ts'],
+    },
```
72636c, line 25: Would it be too heavy handed for `Jest.mergePreset` to provide the root defaults to each project, like e.g.

```typescript
projects: projects.map(project => ({ transform, moduleNameMapper, ...project })
```

samchungy: Yeah I wasn't sure, I think it makes sense though

src/api/jest/index.ts
```diff
@@ -34,6 +34,25 @@ type DefaultOptions =
...
+      return project;
+    }
+
+    return {
+      transform: jestPreset.transform,
+      moduleNameMapper: jestPreset.moduleNameMapper,
+      ...project,
+    };
+  }),
+});
```
72636c, line 56: Thoughts on the following options:

```suggestion
>): Config.InitialOptions => {
  const root = mergeRaw(jestPreset, options);

  return {
    ...root,

    projects: projects?.map((project) => {
      if (typeof project === 'string') {
        return project;
      }

      return {
        ...root,
        ...project,
      };
    }),
  };
};
```

```suggestion
>): Config.InitialOptions => {
  const root = mergeRaw(jestPreset, options);

  return {
    ...root,

    projects: projects?.map((project) => {
      if (typeof project === 'string') {
        return project;
      }

      return {
        moduleNameMapper: root.moduleNameMapper,
        transform: root.transform,
        ...project,
      };
    }),
  };
};
```

Though curiously the [documentation](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig) states the following; is this is meant to work OOTB, or potentially only for the `string` option?

> With the `projects` option enabled, Jest will copy the root-level configuration options to each individual child configuration during the test run, resolving its values in the child's context. This means that string tokens like `<rootDir>` will point to the _child's root directory_ even if they are defined in the root-level configuration.

jest.config.ts
```diff
@@ -4,9 +4,25 @@ export default Jest.mergePreset({
...
   setupFiles: ['<rootDir>/jest.setup.ts'],
-  testPathIgnorePatterns: [
-    '<rootDir>/template/',
-    '\\.int\\.test\\.ts',
-    '/test\\.ts',
+  testPathIgnorePatterns: ['<rootDir>/template/', '/test\\.ts'],
+  watchPathIgnorePatterns: [
+    '<rootDir>/integration/format/',
+    '<rootDir>/integration/lint/',
+  ],
```
72636c, line 11: 🙇

72636c (approved):
🙇

src/api/jest/index.ts
samchungy: Going with the second. With the first you get alot of warnings complaining about unknown config so it appears only a subset of settings actually work for the `projects` key

samchungy: image

Just screenshot for my memory if we ever wonder why again

PR #1305, opened 2023-10-27
Conditionally commit .npmrc in autofix

samchungy:
Conditionally commit `.npmrc` based on the presense of the auth token.

If we want to manage .npmrc, we need some sort of mechanism to be able to auto-push it or else we'll be manually running the `yarn format` 😢


src/api/git/getChangedFiles.int.test.ts
```diff
@@ -78,10 +80,15 @@ it('should exclude file changes based on ignore parameter', async () => {
 
   await Promise.all([
     fs.promises.rm('a'),
-    fs.promises.writeFile('b', '2'),
-    fs.promises.writeFile('c', '2'),
+    fs.promises.writeFile('b', '21'),
```
samchungy, line 83: Not sure if bug but if the size of the contents remains the same it doesn't count as modified :\

AaronMoat:
https://github.com/seek-oss/skuba/pull/1413

PR #1306, opened 2023-10-28
Gracefully Shutdown Node Server

samchungy:
https://aws.amazon.com/blogs/containers/graceful-shutdowns-with-ecs/

Below is a simple test of calling an axios client calling a server repeatedly 3 times. The server has a simulated delay of 5 seconds before returning.

### Before:

1. We are not terminating the server which means our server stays open until ECS kills it. This means that those with a keep alive connections can sometimes get abruptly cut off causing a socket close error or a 502.

image

### After:

1. Server completes the first request
2. The next request successfully reuses the same socket and returns with a 'close' connection header.
3. The http client attempts to create a new socket connection and is met with unable to connect. On AWS, the ALB would route this to a difference instance.

image

This should resolve the remaining proxyStatus=502 present in logs.

#### Caveats

Looks like unfortunately [node.js doesn't close keep alive connections until the keepAliveTimeout ends](https://github.com/nodejs/node/issues/2642) so we are setting the response header for [Connection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection) to 'Close' which will tell clients to stop sending requests to the socket. In my testing, browser http clients and http clients such as Axios, node-fetch and native fetch all respect this and do not attempt to send another request down the socket which allows node.js to close it. I prefer this over forcefully closing the socket and leaving a client with a socket hang up error.

In the event that consumers keep sending requests down the socket, eventually the ECS timeout will be reached and will send a SIGKILL which will bring down the server anyway. I think that this is acceptable.


samchungy:
I'll pop out an autofix a little later

samchungy:
Hmmmm I'm stuck. In my local testing, if I keep requesting the keepAliveTimeout keeps the socket alive as it is actively refreshed if a request comes in, meaning that the server doesn't actually close.

image

samchungy:
Okay fixed by setting the connection header to closed. Most, if not all clients setting the keep alive header should be aware of how to handle close.

template/express-rest-api/src/listen.ts
```diff
@@ -21,3 +21,17 @@ const listener = app.listen(config.port, () => {
 // https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
 // AWS recommends setting an application timeout larger than the load balancer
 listener.keepAliveTimeout = 31000;
+
+// Gracefully shutdown the server
```
72636c, line 25:
```suggestion
// Gracefully shut down the server
```

72636c (approved):
🙌

samchungy:
Going to let this stew while I monitor how it does with Hirer Graph https://github.com/SEEK-Jobs/hirer-graphql-api/pull/2537

AaronMoat:
@samchungy is it looking like this is something we want in?

samchungy:
Unsure if this made a difference TBH

Issue #1311, opened 2023-11-09
Extend lambda cdk template


samchungy:
In light of https://www.serverless.com/blog/serverless-framework-v4-a-new-model

We could probably upgrade this template to include an equivalent canary deployment. I believe there is also a datadog plugin for CDK too that we could use

samchungy:
https://github.com/DataDog/datadog-cdk-constructs/issues/232#issue-2010819696

Some questions to answer

samchungy:
dd trace has issues with bundlers:

https://github.com/DataDog/dd-trace-js?tab=readme-ov-file#bundling

samchungy:
https://github.com/DataDog/dd-trace-js/issues/2160#issuecomment-1175553622

We can use external in the meantime

samchungy:
aws CDK uses the cli esbuild so we can't pass in plugins

https://github.com/aws/aws-cdk/issues/18470

PR #1321, opened 2023-11-17
Introduce bundling with esbuild, `--hotswap` and `--watch`

samchungy:
Related: https://github.com/seek-oss/skuba/issues/1311


template/lambda-sqs-worker-cdk/infra/__snapshots__/appStack.test.ts.snap
```diff
@@ -163,7 +163,7 @@ exports[`returns expected CloudFormation stack for dev 1`] = `
           },
         },
         "FunctionName": "serviceName",
-        "Handler": "app.handler",
+        "Handler": "index.handler",
```
samchungy, line 166: >The handler value will be automatically prefixed with the bundled output file name, index., unless the handler value contains a . character, in which case the handler value is used as-is to allow for values needed by some Lambda extensions.

template/lambda-sqs-worker-cdk/infra/appStack.ts
```diff
@@ -52,11 +53,14 @@ export class AppStack extends Stack {
...
+    const worker = new aws_lambda_nodejs.NodejsFunction(this, 'worker', {
       architecture: aws_lambda.Architecture[architecture],
-      code: new aws_lambda.AssetCode('./lib'),
+      entry: './src/app.ts',
+      bundling: {
+        sourceMap: true,
+        target: 'node20',
+      },
       runtime: aws_lambda.Runtime.NODEJS_20_X,
-      handler: 'app.handler',
```
samchungy, line 59: >Alternatively, an entry file and handler can be specified:

new nodejs.NodejsFunction(this, 'MyFunction', {
  entry: '/path/to/my/file.ts', // accepts .js, .jsx, .cjs, .mjs, .ts, .tsx, .cts and .mts files
  handler: 'myExportedFunc', // defaults to 'handler'
});

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs-readme.html#configuring-esbuild

template/lambda-sqs-worker-cdk/infra/__snapshots__/appStack.test.ts.snap
72636c: I wonder if we should rename the source file to `index.ts` to align 🤔

samchungy: I don't think it's very necessary, just a bundling thing tbh

template/lambda-sqs-worker-cdk/package.json
```diff
@@ -4,6 +4,8 @@
   "scripts": {
     "build": "skuba build",
```
72636c, line 5: Does this render `skuba build` redundant for this template?

template/lambda-sqs-worker-cdk/package.json
```diff
@@ -4,6 +4,8 @@
...
     "deploy": "cdk deploy appStack --require-approval never --context stage=${ENVIRONMENT}",
+    "deploy:hotswap": "yarn deploy --hotswap",
+    "deploy:watch": "yarn deploy:hotswap --watch",
     "format": "skuba format",
     "lint": "skuba lint",
     "package": "yarn install --ignore-optional --ignore-scripts --modules-folder ./lib/node_modules --non-interactive --offline --production",
```
72636c, line 11: Do we still need this packaging malarkey?

72636c (approved):
👏

template/lambda-sqs-worker-cdk/package.json
samchungy: Ruh roh looks like removing this causes issues with test-template

72636c: Heh, I think we can catch and ignore "Command not found" errors there 🙏

72636c (approved):

PR #1327, opened 2023-11-26
Add blue-green deployments to sqs cdk template

samchungy:
Resolves https://github.com/seek-oss/skuba/issues/1311


template/lambda-sqs-worker-cdk/src/app.ts
```diff
@@ -4,7 +4,18 @@ import type { SQSEvent, SQSHandler } from 'aws-lambda';
...
+export const handler: SQSHandler = (event: SQSEvent) => {
+  // Treat an empty object as our smoke test event.
+  if (!Object.keys(event).length) {
+    logger.debug('Received smoke test request');
+    return smokeTest();
+  }
 
-export const handler: SQSHandler = (_: SQSEvent) => {
   logger.info('Hello World!');
+  return;
```
samchungy, line 20: This complained to me about "not all codepaths return a value"

template/lambda-sqs-worker-cdk/infra/appStack.ts
```diff
@@ -51,32 +54,129 @@ export class AppStack extends Stack {
...
+
+    const postHook = new aws_lambda_nodejs.NodejsFunction(
+      this,
+      'worker-post-hook',
+      {
+        ...defaultWorkerConfig,
+        entry: './src/postHook.ts',
+        timeout: Duration.seconds(30),
+        bundling: defaultWorkerBundlingConfig,
+        functionName: 'serviceName-post-hook',
```
samchungy, line 133:
```suggestion
        functionName: '<%- serviceName %>-post-hook',
```

template/lambda-sqs-worker-cdk/infra/appStack.ts
```diff
@@ -51,32 +54,129 @@ export class AppStack extends Stack {
       encryptionMasterKey: kmsKey,
     });
 
+    topic.addSubscription(new aws_sns_subscriptions.SqsSubscription(queue));
```
72636c, line 57: Interesting, this stack deploys the _source_ topic. That feels a bit strange as typically you'd subscribe to an external source 😅

template/lambda-sqs-worker-cdk/infra/appStack.ts
```diff
@@ -51,32 +54,129 @@ export class AppStack extends Stack {
...
-        NODE_OPTIONS: '--enable-source-maps',
+        ...defaultWorkerEnvironment,
         ...context.workerLambda.environment,
       },
-      // aws-sdk-v3 sets this to true by default so it is not necessary to set the environment variable
-      awsSdkConnectionReuse: false,
+      // https://github.com/aws/aws-cdk/issues/28237
+      // This forces the lambda to be updated on every deployment
+      // If you do not wish to use hotswap, you can remove the new Date().toISOString() from the description
+      description: `Lambda description - updated at ${new Date().toISOString()}`,
```
72636c, line 96: Would we rather:

1. Encourage proper descriptions, in which case I'd advocate for a `TODO: ` in here
2. Not bother with descriptions, in which case it can just say `Updated at {now}`

template/lambda-sqs-worker-cdk/infra/appStack.ts
```diff
@@ -51,32 +54,129 @@ export class AppStack extends Stack {
...
+    alias.addEventSource(new aws_lambda_event_sources.SqsEventSource(queue));
 
-    topic.addSubscription(new aws_sns_subscriptions.SqsSubscription(queue));
+    const preHook = new aws_lambda_nodejs.NodejsFunction(
+      this,
+      'worker-pre-hook',
+      {
+        ...defaultWorkerConfig,
+        entry: './src/preHook.ts',
+        timeout: Duration.seconds(30),
```
72636c, line 113: We may want a more generous timeout to allow the generous retry configuration to play out:

https://github.com/seek-oss/skuba/blob/6f0555528126d80a3f035c177697eb1e5de8f03d/template/lambda-sqs-worker-cdk/src/preHook.ts#L11-L19

IIRC this is because CodeDeploy can hang if the hook doesn't provide a well-formed response. Alternatively/additionally we could implement a `Promise.race` or similar to ensure that the hook returns before this timeout.

template/lambda-sqs-worker-cdk/src/postHook.ts
```diff
@@ -0,0 +1,150 @@
...
+): Promise<void> => {
+  const [aliases, versions] = await Promise.all([
+    listAliases(functionName),
+    listLambdaVersions(functionName),
+  ]);
+
+  const aliasMap = new Map(
+    aliases
+      .filter((alias) => alias.FunctionVersion)
+      .map((alias) => [alias.FunctionVersion as string, alias]),
```
72636c, line 75: Can we `flatMap` to avoid the type assertion? 🙏

template/lambda-sqs-worker-cdk/src/postHook.ts
```diff
@@ -0,0 +1,150 @@
...
+    )
+    .sort((a, b) => Number(b.Version) - Number(a.Version))
+    .slice(numberToKeep);
+
+  if (!versionsToPrune.length) {
+    console.log('No function versions to prune');
+    return;
+  }
+
+  console.log(`Pruning ${versionsToPrune.length} function versions`);
```
72636c, line 93: Is it worth logging each pruned version for traceability?

72636c (approved):
This is wonderful 👏

template/lambda-sqs-worker-cdk/infra/appStack.ts
samchungy: Yeah I've been playing with this today. Looks like we can create a subscription to a topic which seems more ideal.

image

template/lambda-sqs-worker-cdk/package.json
72636c, line 1: I may have done something wrong but I had to add the following to deploy under pnpm:

```dotenv
public-hoist-pattern[]=esbuild
public-hoist-pattern[]=ts-node
```

Alternatively we could specify these as direct dependencies

samchungy: I think we can get around this one by using skuba node in the command to run CDK?

samchungy: https://github.com/seek-oss/skuba/blob/ec8014bff06bead1c9f1e0fc635e8de7177fd96a/template/lambda-sqs-worker-cdk/cdk.json#L2

esbuild needing to be hoisted sounds right to me

PR #1339, opened 2023-12-12
Improve input validation error response for Zod unions

### LINKED ISSUE #1338: Zod Union Errors are not logged properly
The current [`validate function`](https://github.com/seek-oss/skuba/blob/master/template/koa-rest-api/src/framework/validation.ts#L39) does not log `invalid_union` Zod Error correctly.

Developers cannot address the issue accurately, like which input field is invalid and why.
Relates to https://github.com/SEEK-Jobs/hirer-audit-cloudwatch-publisher/pull/126

e.g. 
```
const IdDescriptionSchema = z.union([
  z.object({
    id: z.string(),
    description: z.string(),
  }),
  z.object({
    id: z.number(),
    summary: z.string(),
  }),
]);

const input = { id: null, description: 'description' }

validate({
  ctx,
  input,
  schema: IdDescriptionSchema
})
```

**Expect output** 
```
{
  "invalidFields": {
    "/id": "Expect number/string, but got null",
  },
  "message": "Input validation failed",
}
```

**But got**
```
{
  "invalidFields": {
    "/": "Invalid input",
  },
  "message": "Input validation failed",
}
```

JasonZzy0528:
Resolves #1338 

## Changes
- Updated `IdDescriptioNSchema` to a union schema
- ~~Updated `InvalidFields` type to `Record<string, string[]>`~~
- Add a condition to handle `invalid_union` Zod Error

See unexpected output from the unit tests in the [first commit](https://github.com/seek-oss/skuba/commit/c574dd142cb9b99a4777c9e47318f26977286138).


template/koa-rest-api/src/framework/validation.test.ts
```diff
@@ -60,13 +66,21 @@ describe('validate', () => {
...
-            "message": "Input validation failed",
-          }
-        `),
+{
+  "invalidFields": {
+    "/description": [
+      "Required",
+    ],
+    "/id": [
+      "Required",
```
samchungy, line 75: Hmmmm. Is this very useful? Should union errors be grouped accordingly by the member they came from?

JasonZzy0528: We could duplicate the error message for each field.

However, the new implementation at least tell us which field is invalid compared to the old implementation 😛
check unit tests in [first commit](https://github.com/seek-oss/skuba/pull/1339/commits/c574dd142cb9b99a4777c9e47318f26977286138)(only union schema changes)

samchungy: What I'm suggesting is instead of aggregating them, could we perhaps split them into:

```
union1: { (union 1 errors }
union2: { union 2 errors }
```
At the moment I'm not sure what combination of fields would make it not error

JasonZzy0528: I don't know if that is possible. 
The `ZodError` does not provide which union it comes from.

samchungy: Don't you have an array of union errors you're mapping over?

https://github.com/seek-oss/skuba/pull/1339/files#diff-4d7c9e9cdc6b4d6ed3ee02ed3b4a0c5aa141aabfa04ab2169909f9fe269ccca4R39

JasonZzy0528: Updated with the union index, please have a look.

template/koa-rest-api/src/framework/validation.ts
```diff
@@ -51,11 +70,13 @@ export const validate = <
 }): Output => {
   const parseResult = schema.safeParse(input);
   if (parseResult.success === false) {
+    const invalidFields: InvalidFields = {};
+    parseInvalidFieldsFromError(invalidFields, parseResult.error);
```
72636c, line 74: Nitpick: can we keep `parseInvalidFieldsFromError` a pure function that constructs its own response rather than it mutating a parameter?

template/koa-rest-api/src/framework/validation.test.ts
```diff
@@ -43,13 +43,21 @@ describe('validate', () => {
...
-          {
-            "invalidFields": {
-              "/id": "Expected string, received null",
-            },
-            "message": "Input validation failed",
-          }
-        `),
+{
+  "invalidFields": {
+    "/id_0": [
```
72636c, line 48: For what it's worth, the initial concept here was to return [JSON Pointers](https://datatracker.ietf.org/doc/html/rfc6901) to the specific part of the document that failed validation.

This change breaks that convention as it introduces its own syntax for unions, and it means that any clients looking to parse the response would have to understand said syntax. That said, I'm fine with moving ahead with this if we feel it's worthwhile, e.g. because in the typical case we just want to be able to tell what went wrong when reading the error message in its entirety, and because it would be rather unusual to have an organic field name that ends with `_X` that this could be confused with.

.changeset/quiet-pants-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+template/koa-rest-api: fix validation error logs with zod union schema error
```
72636c, line 5:
```suggestion
template/koa-rest-api: Improve input validation error response for Zod unions
```

template/koa-rest-api/src/framework/validation.ts
```diff
@@ -26,15 +27,33 @@ import type { Context } from 'src/types/koa';
  * Returns:
  *
  * ```json
- * { "/advertiserId": "advertiserId is required in the URL" }
+ * { "/advertiserId": ["advertiserId is required in the URL"] }
```
72636c, line 30: Do we want to introduce an array in the typical case where there is only one error message?

template/koa-rest-api/src/framework/validation.test.ts
JasonZzy0528: Some confusion may happen if we want to address the order of union error. 
For example, following JSON Pointers with union error, we may have something like `/id/0`, which may confused with [array schema](https://github.com/colinhacks/zod?tab=readme-ov-file#arrays).

We can probably have an unusual postfix, like `_union0` to clear up confusion at a certain level, although it breaks the convention too.

template/koa-rest-api/src/framework/validation.ts
JasonZzy0528: Will revert the change since we are handling the union error with the index now.

template/koa-rest-api/src/framework/validation.ts
```diff
@@ -29,13 +31,31 @@ import type { Context } from 'src/types/koa';
...
+  for (const [issue, unionIdx] of flattenedIssues) {
+    if (issue.code === ZodIssueCode.invalid_union) {
+      issue.unionErrors.forEach((err, idx) => {
+        err.errors.forEach((childIssue) => {
+          flattenedIssues.push([childIssue, idx]);
+        });
+      });
+    } else {
+      const path = `/${issue.path.join('/')}${
+        unionIdx !== undefined ? `_union${unionIdx}` : ''
```
72636c, line 50: Optional suggestion: while we can't really uphold JSON Pointer compatibility anymore with this approach, perhaps using the same [initial escape character](https://datatracker.ietf.org/doc/html/rfc6901#section-3) would be more immediately recognisable than an underscore, which is relatively common in JSON property names:

```suggestion
        unionIdx !== undefined ? `~union${unionIdx}` : ''
```

template/koa-rest-api/src/framework/validation.ts
```diff
@@ -29,13 +31,31 @@ import type { Context } from 'src/types/koa';
...
+
+const parseInvalidFieldsFromError = ({ errors }: z.ZodError) => {
+  const invalidFields: InvalidFields = {};
+  const flattenedIssues: Array<[z.ZodIssue, UnionErrorIdx]> = errors.map(
+    (issue) => [issue, undefined],
   );
 
+  for (const [issue, unionIdx] of flattenedIssues) {
+    if (issue.code === ZodIssueCode.invalid_union) {
+      issue.unionErrors.forEach((err, idx) => {
```
72636c, line 43: Does this work for nested unions?

JasonZzy0528: Updated in the [commit](https://github.com/seek-oss/skuba/pull/1339/commits/4402bc80463800e396e5299f54bdb1c5c05aa330)

JasonZzy0528: @72636c I have added the logic to handle nested unions. Could you please have a look?🙏

template/koa-rest-api/src/testing/types.ts
```diff
@@ -8,7 +8,13 @@ export type IdDescription = z.infer<typeof IdDescriptionSchema>;
 export const IdDescriptionSchema = z.union([
   z.object({
     id: z.string(),
-    description: z.string(),
+    description: z.union([
+      z.string(),
+      z.object({
+        fontSize: z.number(),
+        content: z.string(),
+      }),
```
72636c, line 16: Sorry for the delay @JasonZzy0528! There doesn't appear to be a nested union here though; do you have a newer version of this file locally?

JasonZzy0528: hmm...I am a bit confused, the `IdDescription` is a union and its `description` field is a union, is `IdDescription` a nested union?

72636c: Ah I misread, sorry about that 😅

template/koa-rest-api/src/framework/validation.ts
72636c: 🙇

I tried re-implementing the function to understand how it worked, particularly as I was about nervous about `push`ing into an array that was being iterated on. I ended up looking at the following diff; do you think this could be useful to more closely reflect the shape of the Zod schema, i.e. where a union sits "before" the `description` property?

```diff
Snapshot name: `validate blocks invalid nested union prop 1`

-     "/description~union0/content~union1": "Required",
-     "/description~union0~union0": "Expected string, received object",
-     "/id~union0": "Expected string, received null",
-     "/id~union1": "Expected number, received null",
-     "/summary~union1": "Required",
+     "~union0/description~union0": "Expected string, received object",
+     "~union0/description~union1/content": "Required",
+     "~union0/id": "Expected string, received null",
+     "~union1/id": "Expected number, received null",
+     "~union1/summary": "Required",
```

```typescript
const parseInvalidFieldsFromError = (err: z.ZodError): InvalidFields =>
  Object.fromEntries(parseTuples(err, {}));

const parseTuples = (
  { errors }: z.ZodError,
  unions: Record<number, number[]>,
): Array<[string, string]> =>
  errors.flatMap((issue) => {
    if (issue.code === ZodIssueCode.invalid_union) {
      return issue.unionErrors.flatMap((err, idx) =>
        parseTuples(err, {
          ...unions,
          [issue.path.length]: [...(unions[issue.path.length] ?? []), idx],
        }),
      );
    }

    const path = ['', ...issue.path]
      .map((prop, idx) =>
        [
          prop,
          ...(unions[idx] ?? []),
        ].join('~union'),
      )
      .join('/');

    return [[path, issue.message]] as const;
  });
```

Alternatively I can get it to behave identically to yours with a few tweaks:

```diff
const parseTuples = (
  { errors }: z.ZodError,
  unions: Record<number, number[]>,
): Array<[string, string]> =>
  errors.flatMap((issue) => {
    if (issue.code === ZodIssueCode.invalid_union) {
      return issue.unionErrors.flatMap((err, idx) =>
        parseTuples(err, {
          ...unions,
          [issue.path.length]: [...(unions[issue.path.length] ?? []), idx],
        }),
      );
    }

-    const path = ['', ...issue.path]
+    const path = issue.path
      .map((prop, idx) =>
        [
          prop,
          ...(unions[idx] ?? []),
+         ...((idx + 1 === issue.path.length && unions[idx + 1]) || []),
        ].join('~union'),
      )
      .join('/');

-   return [[path, issue.message]] as const;
+   return [[`/${path}`, issue.message]] as const;
  });
```

JasonZzy0528: Actually, I think your approach output is more readable.
Would you like to push your changes? or I can update and push.

72636c: Happy for you to decide which is best here and push 👍. Thanks for your patience!

72636c (approved):

template/koa-rest-api/src/framework/validation.ts
JasonZzy0528: Updated

ConradLang:
@samchungy , @72636c : I find the `~union123` in the path confusing from a consumer point of view.

It's hard to easily understand why the object I sent is invalid looking at this:

```json
{
  "invalidFields": {
    "~union0/description~union0": "Required",
    "~union0/description~union1": "Required",
    "~union0/id": "Expected string, received null",
    "~union1/id": "Expected number, received null",
    "~union1/summary": "Required"
  },
  "message": "Input validation failed"
}
```

- Even as the code maintainer, I'd go to the schema and my editor could highlight all occurrences of the property.
- The `~union123` suffix doesn't convey additional information i.e. which union contains the validation rule.

A simpler error is far easier to read and understand:

```json
{
  "invalidFields": {
    "/description": "Required",
    "/id": [
      "Expected string, received null",
      "Expected number, received null"
    ],
    "/summary": "Required"
  },
  "message": "Input validation failed"
}
```

I'd suggest we change the validation object to be simpler ... even if it has an array of validation error messages per path.

72636c:
@ConradLang the challenge isn't really around having multiple messages per path, it's that errors can correspond to a particular branch of a union that can occur at any arbitrary level of nesting in the object schema.

This lets us communicate mutually exclusive props / validation:

```json
{
  "~union0/xor1": "Required",
  "~union1/xor2": "Required"
}
```

Whereas this would be wrong:

```json
{
  "xor1": "Required",
  "xor2": "Required"
}
```

ConradLang:
Could we at least have the option to omit the unions as I believe that it's not very readable or useful unless you can infer the union type ... and even then I'd prefer to have a clean path name.

72636c:
The template is just a starting point. I could go either way on what makes sense as the default, but you can just delete the union handling code when you create a new Koa project.

PR #1341, opened 2023-12-13
Use `--env-file` to simplify self-development

72636c:
Setting this to run `nr test*` on each new terminal session was getting annoying.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+NODE_OPTIONS=--experimental-vm-modules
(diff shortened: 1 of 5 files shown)


samchungy:
Could we use https://nodejs.org/api/cli.html#--env-fileconfig

72636c:
@samchungy Requires Node.js 20 but I reckon that's fine for local dev 👍

samchungy (approved):

PR #1346, opened 2023-12-14
RFC: Don't silently write on lint

AaronMoat:
Currently, if .gitignore/other ignore files need to be fixed by `format`, you can get away with never having them fixed because lint never fails. I've seen cases where they're not pushed with autofix (unsure if always or just if no write perms?)

There's way more work to do here (same goes for skuba upgrade), autofix would be nice, but hoping to get feedback on direction before committing 

Fairly important to get right before pnpm / managing npmrc

### DIFF
--- a/src/cli/lint/index.ts
+++ b/src/cli/lint/index.ts
@@ -13,7 +13,7 @@ export const lint = async (
   tscOutputStream: Writable | undefined = undefined,
   workerThreads = true,
 ) => {
-  await Promise.all([tryRefreshIgnoreFiles(), upgradeSkuba()]);
+  await Promise.all([tryRefreshIgnoreFiles({ linting: true }), upgradeSkuba()]);
 
   const opts: Input = {
     debug: hasDebugFlag(args),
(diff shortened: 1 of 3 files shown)


src/cli/lint/index.ts
```diff
@@ -13,7 +13,7 @@ export const lint = async (
   tscOutputStream: Writable | undefined = undefined,
   workerThreads = true,
 ) => {
-  await Promise.all([tryRefreshIgnoreFiles(), upgradeSkuba()]);
+  await Promise.all([tryRefreshIgnoreFiles({ linting: true }), upgradeSkuba()]);
```
72636c, line 16: Sorry for the late review, here are my thoughts:

1. Totally agree that we need to be careful about automatically committing `.npmrc` changes
2. I'd definitely want to retain ignore file autofixes on the happy path. Another option may be to revisit whether we should fail the command if autofix fails, or perhaps if autofix fails _and_ includes a refreshed ignore file.

72636c: There's a chance that this could tie in to https://github.com/seek-oss/skuba/pull/1274#discussion_r1424898271; maybe patches should declare whether they are mandatory, in which case they should fail the build if not autofixed?

PR #1350, opened 2023-12-18
RFC: Remove `BUILDPLATFORM`

72636c:
I'm pretty split on this one. It should be largely inconsequential for the general case where an engineer is running Apple Silicon locally and building for a Graviton environment, but it has some implications when these platforms are not aligned.

https://docs.docker.com/build/guide/multi-platform/#platform-build-arguments

https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

---

This is my understanding:

Preferring `BUILDPLATFORM`:

- Quicker to build when local platform differs from deployment target
- Local behaviour is more likely differ from CI
- Locally built image may not be deployable during disaster recovery

Preferring to hardcode `amd64` or `arm64`:

- Slower to build when local platform differs from deployment target
- Local behaviour is less likely to differ from CI
- Locally built image should be deployable during disaster recovery
- More explicit in the Dockerfile

Preferring no `--platform` / `TARGETPLATFORM`:

- Slower to build when local platform differs from deployment target
- Local behaviour is less likely to differ from CI
- Locally built image should be deployable during disaster recovery
- Somewhat implicit and depends on tooling support for the [`--platform` build option](https://docs.docker.com/engine/reference/commandline/build/#options)


samchungy (approved):

72636c:
Came back around to this and landed on the same conclusion, so I'll get this merged once I write up a changeset, then look at improving guidance in https://seek-oss.github.io/skuba/docs/deep-dives/arm64.html for existing x86 projects.

PR #1357, opened 2023-12-20
Disable tsconfig baseUrl

samchungy:
This one annoys me when it comes to auto importing for this project. If we turn off baseUrl, the IDE will always import relatively.


72636c (approved):
Well that was too simple to be true! 🧹

Issue #1380, opened 2024-01-09
Patch Distroless images to `-debian12`


72636c:
https://github.com/GoogleContainerTools/distroless/tree/main/nodejs#image-contents

---

Prior art: https://github.com/seek-oss/skuba/blob/db7eae8d18bf23617fd4b3800c0fce7ef610f313/src/cli/configure/upgrade/patches/7.3.1/patchDockerfile.ts

PR #1381, opened 2024-01-12
Patch to debian12

AaronMoat:
Resolves https://github.com/seek-oss/skuba/issues/1380

### DIFF
--- a/src/cli/configure/upgrade/patches/7.3.1/patchDockerfile.ts
+++ b/src/cli/configure/upgrade/patches/7.3.1/patchDockerfile.ts
@@ -8,8 +8,8 @@ import { createDestinationFileReader } from '../../../analysis/project';
 
 const DOCKERFILE_FILENAME = 'Dockerfile';
 
-const NON_DEBIAN_REGEX = /gcr.io\/distroless\/nodejs:(16|18|20)/g;
-const DEBIAN_REGEX = /gcr.io\/distroless\/nodejs(16|18|20)-debian11/g;
+const NON_DEBIAN_REGEX = /gcr.io\/distroless\/nodejs:(18|20)/g;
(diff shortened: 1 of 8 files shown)


src/cli/configure/upgrade/patches/7.3.1/patchDockerfile.ts
```diff
@@ -8,8 +8,9 @@ import { createDestinationFileReader } from '../../../analysis/project';
 
 const DOCKERFILE_FILENAME = 'Dockerfile';
 
-const VERSION_REGEX = /gcr.io\/distroless\/nodejs:(16|18|20)/g;
-const VERSION_DEBIAN_REPLACE = 'gcr.io/distroless/nodejs$1-debian11';
+const NON_DEBIAN_REGEX = /gcr.io\/distroless\/nodejs:(16|18|20)/g;
+const DEBIAN_REGEX = /gcr.io\/distroless\/nodejs(16|18|20)-debian11/g;
```
72636c, line 12: It looks like there is no `nodejs16-debian12` support: https://console.cloud.google.com/gcr/images/distroless

AaronMoat: Ooh that's complicated then. Maybe it should be two patches

AaronMoat: Ah, but skuba has engineConfig requiring 18 so this won't run on 16 anyway - maybe we just do the 18/20

72636c: Oh nice one, sounds good 👍

72636c (approved):

PR #1392, opened 2024-01-17
Propagate environment #1358

### LINKED ISSUE #1358: Use propagate-environment: true in Docker Compose config
https://github.com/buildkite-plugins/docker-compose-buildkite-plugin?tab=readme-ov-file#pipeline-variables

Suggested by @ajain58

This may simplify the overall configuration where it can be a little confusing to realise you need to update docker-compose.yml to propogate more environment variables. This also simplifies our [GitHub Autofixes](https://seek-oss.github.io/skuba/docs/deep-dives/github.html#github-autofixes) functionality where we require users to configure environment variables. If we ever decide we need more environment variables we won't need a user to configure extra environment variables manually.

Things to check:
1. If you already have an environment specified in the docker-compose.yml does this override or add to that array? Or does that config override this one?
2. Does it propagate only the agent's predefined environment variables, or does it propagate ones we fetch dynamically, eg the NPM_TOKEN with the ssm plugin

Things that would need to be updated:

1. Templates
2. Documentation surrounding BuildKite environment variables we need to propagate.

ajain58:
Adds `propagate-environment: true` to all pipeline templates and removes the environment definition from docker compose files

Closes #1358


template/express-rest-api/docker-compose.yml
```diff
@@ -2,16 +2,6 @@ version: '3.7'
...
-    environment:
-      # Enable Buildkite + GitHub integrations.
-      - BUILDKITE
-      - BUILDKITE_AGENT_ACCESS_TOKEN
-      - BUILDKITE_BRANCH
-      - BUILDKITE_BUILD_NUMBER
-      - BUILDKITE_JOB_ID
-      - BUILDKITE_PIPELINE_DEFAULT_BRANCH
-      - BUILDKITE_STEP_ID
-      - GITHUB_API_TOKEN
```
72636c, line 14: `GITHUB_API_TOKEN` isn't a built-in Buildkite environment variable, so it may still need to be specified unless your testing shows otherwise:

https://github.com/search?q=org%3ASEEK-Jobs%20propagate-environment%20GITHUB_API_TOKEN&type=code

template/lambda-sqs-worker/docker-compose.yml
```diff
@@ -2,21 +2,6 @@ version: '3.7'
...
-      - BUILDKITE_BUILD_NUMBER
-      - BUILDKITE_JOB_ID
-      - BUILDKITE_PIPELINE_DEFAULT_BRANCH
-      - BUILDKITE_STEP_ID
-      - GITHUB_API_TOKEN
-      # Tag AWS resources with the commit hash.
-      - BUILDKITE_COMMIT
-      # Pass through application configuration.
-      - ENVIRONMENT
-      - VERSION
```
72636c, line 19: Same here

docs/deep-dives/github.md
```diff
@@ -43,18 +43,12 @@ steps:
 ```
 
 With Docker Compose,
-declare the environment variables and volume mounts in your [Compose file]:
+declare volume mounts in your [Compose file].
+You won't need to declare the environment variables in the compose file anymore since `propagate-environment: true` will ensure the variables defined are propagated to the docker environment.
```
72636c, line 47: Can we document how a step with the `docker-compose` plugin should look like, similar to the example above for the `docker` plugin?

72636c (commented):
Great find. Looks like a straightforward port of the Docker plugin implementation https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/344 so I'm not seeing any additional risks.

docs/deep-dives/github.md
samchungy, line 47: Could we include a similar buildkite pipeline example in the same vein as the docker plugin section above?

template/express-rest-api/docker-compose.yml
samchungy: image

Looks like it fails

ajain58: I've added this to the pipeline environments so that it flows through to docker as well, just like the examples from this search: https://github.com/search?q=org%3ASEEK-Jobs%20propagate-environment%20GITHUB_API_TOKEN&type=code

ajain58: ☝️ think defining it on the pipeline with the `propagate-environment` should fix this and we won't need to still pass the environment explicitly in the `docker-compose` anymore.

72636c: Nice one, I like the symmetry between our `docker` and `docker-compose` usage 👍

template/lambda-sqs-worker/.buildkite/pipeline.yml
```diff
@@ -65,6 +66,9 @@ steps:
       - *docker-ecr-cache
       - docker-compose#v4.16.0:
           run: app
+          environment:
+            - GITHUB_API_TOKEN
```
72636c, line 70:
```suggestion
            - ENVIRONMENT
            - GITHUB_API_TOKEN
            - VERSION
```

template/lambda-sqs-worker-cdk/.buildkite/pipeline.yml
```diff
@@ -56,6 +57,9 @@ steps:
       - *docker-ecr-cache
       - docker-compose#v4.16.0:
           run: app
+          environment:
+            - GITHUB_API_TOKEN
```
72636c, line 61:
```suggestion
            - ENVIRONMENT
            - GITHUB_API_TOKEN
            - VERSION
```

docs/deep-dives/github.md
```diff
@@ -43,8 +43,33 @@ steps:
 ```
 
 With Docker Compose,
-declare the environment variables and volume mounts in your [Compose file]:
+declare volume mounts in your [Compose file].
 
+You won't need to declare the environment variables in the compose file since `propagate-environment: true` will ensure the variables defined are propagated to the docker environment.
+
+For example, with the [Docker Buildkite plugin], we will need to add the `propagate-environment: true`:
```
72636c, line 50:
```suggestion
For example, with the [Docker Compose Buildkite plugin]:
```

docs/deep-dives/github.md
```diff
@@ -43,8 +43,33 @@ steps:
...
+      GET_GITHUB_TOKEN: 'please'
+    plugins:
+      - docker#v5.0.0:
+          # Enable GitHub integrations.
+          environment:
+            - GITHUB_API_TOKEN
+          propagate-environment: true
+          volumes:
+            # Mount cached dependencies.
+            - /workdir/node_modules
```
72636c, line 67:
```suggestion
      - *aws-sm
      - *private-npm
      - *docker-ecr-cache
      - docker-compose#v4.16.0:
          # Enable GitHub integrations.
          environment:
            - GITHUB_API_TOKEN
          run: app
          propagate-environment: true
```

72636c (commented):
👏

We should add a changeset for this

docs/deep-dives/github.md
ajain58: what should this link to? there is no `[Docker Compose Buildkite plugin]` doco linked down here: https://github.com/seek-oss/skuba/blob/master/docs/deep-dives/github.md?plain=1#L123

72636c: I think here: https://github.com/buildkite-plugins/docker-compose-buildkite-plugin

ajain58: updated

72636c:
@ajain58 Great question, we have an informal & impure 🙃 practice of filing template updates as `patch`es. This is dishonest to semver but the rationale is that existing projects are the ones that upgrade to new versions & review release notes, and once you've run `skuba init` then updates to the built-in templates don't really impact you beyond an FYI about about a best current practice.

72636c:
That reminds me that we should update https://github.com/seek-oss/skuba/blob/master/CONTRIBUTING.md#creating-a-changeset

docs/deep-dives/github.md
```diff
@@ -43,7 +43,37 @@ steps:
...
+```
+
+And then your [Compose file] will not need to define the `environment` like you had to before to ensure the environment variables are available in Docker
+
+Docker compose file before:
```
samchungy, line 76: Hmmm. I don't think we have to document the before and after within the deep-dive itself. We can probably get away with just showing what the optimal configuration is.

We can probably move some of the `before` and `after` scenarios to the changeset

template/lambda-sqs-worker-cdk/.buildkite/pipeline.yml
```diff
@@ -56,6 +61,11 @@ steps:
       - *docker-ecr-cache
       - docker-compose#v4.16.0:
           run: app
+          environment:
+            - ENVIRONMENT
+            - GITHUB_API_TOKEN
+            - VERSION
```
samchungy, line 67: Any ideas where `VERSION` comes from? @72636c. I can see `ENVIRONMENT` in the UI so it will get propagated successfully by the `propagate-environment` setting.

samchungy: https://github.com/ajain58/skuba/blob/bbdd1782c27d400ec0abc4908246ee206fd4d6dd/template/lambda-sqs-worker/.buildkite/pipeline.yml#L47-L48

Is this template meant to have this line?

72636c: Ah yes, I think that's right re: `VERSION`. Maybe we should extend the CDK template to use the version variable too

CDK

https://github.com/seek-oss/skuba/blob/454710c4ab068c073918ec2426896756169ce22d/template/lambda-sqs-worker-cdk/src/app.ts#L4-L6

Serverless

https://github.com/seek-oss/skuba/blob/454710c4ab068c073918ec2426896756169ce22d/template/lambda-sqs-worker/src/config.ts#L48

https://github.com/seek-oss/skuba/blob/454710c4ab068c073918ec2426896756169ce22d/template/lambda-sqs-worker/src/framework/logging.ts#L14-L17

samchungy:
I did a quick test and pipeline variables which are defined within steps and/or on the pipeline at upload time also get passed to the container via the `propagate-environment` option.

72636c (approved):
Thanks a bunch for this 🙌

PR #1398, opened 2024-01-19
RFC: Fix CI

72636c:
Prettier has gained support for honouring trailing comma configuration in JSONC files and is applying this to our `tsconfig.json`s.

I'm not entirely convinced by this change, and could be persuaded to mess with our Prettier adapter and toggle trailing commas off for `tsconfig.json` in particular. Notably this is not affecting our `tsconfig.build.json` which could appear inconsistent.


AaronMoat (approved):
In theory this needs a changeset

AaronMoat (approved):

Issue #1286, opened 2023-10-09
Manage .npmrc


samchungy:
We need to manage the repository .npmrc file in order to be able to hoist some of skuba's internal dependencies.

```
# managed by skuba
public-hoist-pattern[]="*eslint*"
public-hoist-pattern[]="*prettier*"
public-hoist-pattern[]="@types*"
public-hoist-pattern[]="*jest*"
public-hoist-pattern[]="tsconfig-seek"
# end managed by skuba
```

This presents a couple of issues, so in our upgrade script we need:

1. We need to modify the .dockerIgnore file, .gitIgnore file to no longer ignore .npmrc. We could potentially do this based on the existence of a pnpm lock file

2. If a workspace has an existing .npmrc file with an auth token we should discard it and place our .npmrc in its place. If it has existing settings we can place it below our managed section.

72636c:
I think the last thing to cover off for the [next version](https://github.com/seek-oss/skuba/pull/1329#issuecomment-1859528370) is to ensure

> Ignore file refreshing and skuba configure do not expose or otherwise mangle existing private .npmrcs

We should review if this comment is still relevant:

https://github.com/seek-oss/skuba/blob/47653f749e0bde9f5539cba3d83cb43d6dc5f4d9/src/cli/lint/internalLints/refreshIgnoreFiles.ts#L25-L29

PR #1413, opened 2024-02-04
Manage .npmrc

AaronMoat:
Resolves https://github.com/seek-oss/skuba/issues/1286

Alternative take / revival of https://github.com/seek-oss/skuba/pull/1305/ after the new linting system is in

Should this be major? Might it break CI?


AaronMoat:
Git isn't picking up that refreshIgnoreFiles.test.ts was a rename - can see the changes in the first commit though.

AaronMoat:
Tests failing due to https://github.com/seek-oss/skuba/pull/1414

src/cli/lint/internalLints/refreshConfigFiles.ts
```diff
@@ -12,37 +12,47 @@ import { createDestinationFileReader } from '../../configure/analysis/project';
...
+  s
+    .split('\n')
+    .filter((line) => !line.includes('authToken'))
+    .join('\n');
+
+const REFRESHABLE_CONFIG_FILES = [
+  { name: '.eslintignore' },
+  { name: '.gitignore', additionalMapping: ensureNoNpmrcExclusion },
+  { name: '.prettierignore' },
+  { name: '.npmrc', additionalMapping: ensureNoAuthToken },
```
72636c, line 31: Would it make sense to make this refresh/management conditional on the package manager? https://github.com/seek-oss/skuba/blob/ed2e36ad22498640977da1216fe46388837f338d/src/utils/packageManager.ts#L37-L39

I think that would rule out any potential backward incompatibilities with Yarn projects, which are still the overwhelming majority in SEEK-Jobs.

AaronMoat: Pushed an attempt - probably needs scrutiny

.npmrc
```diff
@@ -2,7 +2,9 @@
 public-hoist-pattern[]="@types*"
 public-hoist-pattern[]="*eslint*"
 public-hoist-pattern[]="*prettier*"
+public-hoist-pattern[]="esbuild"
 public-hoist-pattern[]="tsconfig-seek"
+public-hoist-pattern[]="jest"
```
72636c, line 7: Sorry 😝

```suggestion
public-hoist-pattern[]="jest"
public-hoist-pattern[]="tsconfig-seek"
```

src/cli/lint/internalLints/refreshConfigFiles.ts
```diff
@@ -5,44 +5,83 @@ import { writeFile } from 'fs-extra';
...
+    additionalMapping: (
+      contents: string,
+      packageManager: PackageManagerConfig,
+    ) => {
+      if (packageManager.command === 'pnpm') {
+        return ensureNoNpmrcExclusion(contents);
+      }
+      return contents;
+    },
+  },
```
72636c, line 43: The remaining concern I have is:

1. An existing project like https://github.com/seek-oss/koala/blob/e3278b52be501f891897c5cf2621657369e9b285/.gitignore#L17 uses Yarn and relies on skuba to gitignore `.npmrc`
2. An engineer has copied their `.npmrc` + auth token into the local project directory for testing purposes
3. When this new version applies its config refresh, `mergeWithIgnoreFile` will remove the `.npmrc` gitignore
4. The engineer then commits and pushes their local `.npmrc` + auth token

My initial thought was that this should be _preprocessed_, something like checking if `.npmrc` was previously gitignored and then preserving it at the bottom of the gitignore file, before `mergeWithIgnoreFile` goes and updates the skuba managed section. It could also be implemented as a patch if we can get that to run before the config refresh.

72636c (commented):
🙇 code looks great, I think there's just one small niggle to cover off

.changeset/itchy-avocados-peel.md
```diff
@@ -0,0 +1,8 @@
+---
+'skuba': minor
+---
+
+lint: Manage .npmrc files for `pnpm` installations of skuba
+
+skuba now manages `.npmrc` files when using `pnpm` for compatibility purposes.
+skuba will attempt to ignore authTokens, and remove any `.gitignore` entries that ignore `.npmrc`.
```
72636c, line 8: Nitpick: IIRC Changesets doesn't respect the Markdown spec for line breaks 😞

```suggestion
lint: Manage `.npmrc` for pnpm projects

skuba now manages `.npmrc` when a project uses `pnpm` to enable [dependency hoisting](https://pnpm.io/npmrc#dependency-hoisting-settings). It will attempt to strip sensitive auth tokens from the file, and remove `.gitignore` entries that ignore `.npmrc`.
```

src/cli/lint/internalLints/refreshConfigFiles.ts
AaronMoat: Hmm I did a test. Does it fully cover the scenarios?

I created a test repository, and did the following:

- Set up a `.gitignore` file with `mysecretfile`. Pushed it. Created a `mysecretfile` with a secret in it. 
- In a separate checkout, modified .gitignore to remove that entry, and in the same commit, created a mysecretfile with fake "managed" content. Pushed.
- Pulled in the original checkout, and the new managed content overwrote the file and my secret was deleted.

In this scenario, it would be quite hard to accidentally push the secret. 

The cases where I can see it happening: 

- Not all files are committed (I could see this potentially happening if doing `git commit -am` not picking up new files). I suppose leaving it up to manual intervention helps this somewhat, eg. by adding a "# TODO: remove this line and make sure you push the npmrc"?
- If for some reason we don't write npmrc and gitignore (e.g. one crashes). This seems unlikely. 

I _am_ open to workshopping this, just want to understand what we're trying to guard against. What are your thoughts?

AaronMoat (commented):
Speaking of not committing all files - autofix will skip changes to npmrc (`AUTOFIX_IGNORE_FILES`), need to fix that (in a safe way)

src/cli/lint/internalLints/refreshConfigFiles.ts
72636c: All good, I may be mistaken but I think the scenario is that

- Bullet two runs in CI so the engineer isn't necessarily looking at what's going on
- Bullet two doesn't synthesise a conflicting `mysecretfile` because the project is still Yarn based

AaronMoat: Ah, I see, of course 🤦 

A way to tackle that I suppose is to just always have the npmrc synthesised? I.e. reverting e951645a72d8b82d1146b05fb38da4c2bdea6a5a - I can't [really see it causing too many issues](https://github.com/search?q=%28org%3Anpm+OR+org%3Ayarnpkg%29+public-hoist-pattern&type=code) with other package managers, but hard to entirely prove

src/cli/lint/autofix.ts
```diff
@@ -93,6 +96,17 @@ const shouldPush = async ({
   return true;
 };
 
+const getIgnores = async (dir: string): Promise<Git.ChangedFile[]> => {
+  const contents = await createDestinationFileReader(dir)('.gitignore');
+
+  // If an .npmrc has secrets, we need to ignore it
+  if (hasNpmrcSecret(contents ?? '')) {
```
72636c, line 103: Should this be:

```suggestion
  const contents = await createDestinationFileReader(dir)('.npmrc');

  // If an .npmrc has secrets, we need to ignore it
  if (hasNpmrcSecret(contents ?? '')) {
```

src/cli/configure/upgrade/patches/7.3.1/moveNpmrcOutOfGitignoreManagedSection.ts
```diff
@@ -0,0 +1,85 @@
...
+import { log } from '../../../../../utils/logging';
+import { createDestinationFileReader } from '../../../analysis/project';
+
+const NPMRC_IGNORE_SECTION = `
+
+# Ignore .npmrc. Previously managed by skuba, but removed as pnpm projects require a managed .npmrc.
+# If migrating to pnpm, remove this line.
+# IMPORTANT: Ensure you add a .npmrc file IN THE SAME COMMIT to avoid secret exposure.
+# skuba can generate an .npmrc for you with \`skuba format\` if you move to pnpm - but committing an empty file is sufficient for security reasons.
+.npmrc
```
72636c, line 16: How about this:

```suggestion
# Ignore .npmrc. This is no longer managed by skuba as pnpm projects use a managed .npmrc.
# IMPORTANT: if migrating to pnpm, remove this line and add an .npmrc IN THE SAME COMMIT.
# You can use \`skuba format\` to generate the file or otherwise commit an empty file.
# This will conflict with a local .npmrc and make it more difficult to unintentionally commit auth secrets.
```

.changeset/itchy-avocados-peel.md
```diff
@@ -0,0 +1,7 @@
+---
+'skuba': minor
+---
+
+lint: Manage `.npmrc` for pnpm projects
+
+skuba now manages `.npmrc` when a project uses `pnpm` to enable [dependency hoisting](https://pnpm.io/npmrc#dependency-hoisting-settings). It will attempt to strip sensitive auth tokens from the file, and remove `.gitignore` entries that ignore `.npmrc`.
```
72636c, line 7: In lieu of a pnpm migration guide, maybe we can leave this like so for the time being:

```suggestion
skuba now manages a section of `.npmrc` when a project uses `pnpm` to enable [dependency hoisting](https://pnpm.io/npmrc#dependency-hoisting-settings). It will continue to avoid committing autofixes to the file if it contains auth secrets.
```

72636c (approved):
This looks wonderful, thank you 🚀

PR #1414, opened 2024-02-04
Revert JSONC on tsconfig.json changes

AaronMoat:
Looks like prettier reverted the change for tsconfig.json in another patch version: https://github.com/prettier/prettier/blob/main/CHANGELOG.md#use-json-parser-for-tsconfigjson-by-default-16012-by-sosukesuzuki


.changeset/great-trains-doubt.md
```diff
@@ -1,23 +0,0 @@
----
-'skuba': patch
----
-
-deps: Prettier 3.2
-
-A notable change in this release is the application of trailing newlines to JSONC files, including `tsconfig.json`:
```
72636c, line 7: Let's keep the changeset, but we can delete L7–22

72636c (approved):
😌

PR #1415, opened 2024-02-04
Propose a release preamble

AaronMoat:
🤔 I am not 100% sure if this will break anything, e.g. if it will propagate the notes to the `releases` section as well (or perhaps entirely break that?)  

Edit: reading the source code for the changesets action, I reckon (but can't promise) it'll be ok

But the [release notes for this release](https://github.com/seek-oss/skuba/pull/1329) are _huge_ and I think adding some important callouts at the top would be nice.


.changeset/.PREAMBLE.md
```diff
@@ -0,0 +1,8 @@
+This version of skuba should not be a large upgrade effort for most use-cases, but does contains some large changes:
+
+- Major upgrades to the linting and patching systems, which may result in some code changes on version upgrades.
+  To make upgrades easy now and going forward, we recommend setting up [GitHub autofixes](https://seek-oss.github.io/skuba/docs/deep-dives/github.html#github-autofixes).
+- New projects will now be initialised with `pnpm`.
+  A future release of `skuba` will likely transition existing projects to `pnpm` as well.
```
72636c, line 6: Still paranoid about changeset formatting, and I figured we should avoid saying "major" on a "minor" release 🙂

```suggestion
This version of skuba should not require significant upgrade effort for most projects, but it does contain some notable changes:

- Internal linting and patching have been overhauled to streamline code generation on version upgrades.

  To make upgrades easy now and going forward, we recommend setting up [GitHub autofixes](https://seek-oss.github.io/skuba/docs/deep-dives/github.html#github-autofixes).

- New projects will now be initialised with pnpm.

  A future release of skuba may transition existing projects to pnpm.
```

72636c (approved):
> if it will propagate the notes to the `releases` section as well (or perhaps entirely break that?)

I think CHANGELOG.md should be authoritative: https://github.com/seek-oss/skuba/pull/1329/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

But happy to be proven wrong in this release, and we can always ninja edit the release note 🙂

PR #1424, opened 2024-02-06
Enter beta mode

AaronMoat:


72636c:
Is it worth trying to push this to `beta` rather than `master`? I'm not sure how gracefully we can exit pre mode.

AaronMoat:
I can probably just run the snapshot on this branch?

PR #1425, opened 2024-02-06
Horrible hacks

AaronMoat:
I'm sorry

---

Trying to get snapshots to public as a pre-release version e.g. next-version-branch-date rather than 0.0.0-branch-date so that patches run only if upgrading.


72636c (approved):
🫣

PR #1432, opened 2024-02-06
Always dedup

AaronMoat:
wip


AaronMoat:
This is probs unnecessary, poses risk given the pattern matching is simple

PR #1433, opened 2024-02-07
Manage .dockerignore

AaronMoat:
- Include in moving npmrc patches out of managed 
- Start managing it


.dockerignore
```diff
@@ -12,7 +12,6 @@ node_modules*/
 /tmp*/
 
 .DS_Store
-.npmrc
```
AaronMoat, line 15: this is the only manual testing I have done 😆

samchungy (approved):

PR #1434, opened 2024-02-07
add .npmrc mounts

samchungy:
Looks like `pnpm fetch` reads the `.npmrc` for settings and outputs a `publicHoistPattern` in `node_modules/.modules.yaml`

image

As a result, when we go to run `pnpm install --offline` it fails silently and asks if we would like to destroy `node_modules`

image

This results in `skuba` not being properly hoisted.


AaronMoat (approved):
Dunno what I'm doing with mounts but seems fine. I don't think we need a changeset as it'd be largely redundant?

samchungy:
> Dunno what I'm doing with mounts but seems fine. I don't think we need a changeset as it'd be largely redundant?

Yeah we haven't release yet, so it should be okay

PR #1441, opened 2024-02-08
Clean up deleteFiles lint

AaronMoat:
I don't think we need it: https://github.com/search?q=org%3ASEEK-Jobs+path:Dockerfile-incunabulum&type=code


72636c (approved):
🙇

PR #1444, opened 2024-02-10
Debug skuba commands hanging

AaronMoat:
image

^ :(


src/skuba.ts
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env node
 
+// @ts-expect-error - no types
+import whyIsNodeRunning from 'why-is-node-running';
```
samchungy, line 4: Can we create a .d.t.s for it instead:

```
declare module 'why-is-node-running' {
  export default function whyIsNodeRunning(): void;
}
```

Should work I thinks

samchungy (approved):

PR #1449, opened 2024-02-17
Allow skuba node to run forever

AaronMoat:
Forgot that skuba node could be used in CI for maintenance-y things


AaronMoat:
Broken due to https://github.com/seek-oss/skuba/pull/1450

samchungy (approved):

PR #1450, opened 2024-02-17
Fix lambda CDK snapshot churn

AaronMoat:
This was pissing me off. The logical ids depend on the code, lock files, etc (e.g. in master get the snapshot passing then change Hello World! to Hello World!! and it'll fail). 

Low value in asserting on specific hash of all the content, so use a stable identifier.


AaronMoat:
😭 no idea why this is different in ci vs local now

samchungy (approved):

PR #1463, opened 2024-03-04
Remove fdir dependency

AaronMoat:
We're getting some peer dependency issues between fdir and picomatch, IMO best to just kill it as we only use it one spot for a narrow scope / fairly simple use case (but also, how is directory crawling not built in?)


tadhglewis (approved):

samchungy (commented):
I suspect @72636c will have opinions on the mutation 😛

72636c (approved):

PR #1468, opened 2024-03-06
refactor(cdk lambda): Renamed stage environment parameter name

jjshinobi:
`SOMETHING` as a stage name in env parameters is confusing. It is not 100% clear it should be replaced when a new project is generated from the template. It should be OK to default it to `STAGE` which reflects exactly what it is. The alternative is to call it `ENVIRONMENT` but "stage" is already used inside infra code so maybe `STAGE` is better.


72636c:
Thanks for the contribution @jjshinobi!

This was really just a placeholder though I worry about making it less obvious to remove if you don't use it. I'd be supportive of this change if the template was updated to actually consume the environment variable as seen in the other templates (that call it `ENVIRONMENT`).

jjshinobi:
> Thanks for the contribution @jjshinobi!
> 
> This was really just a placeholder though I worry about making it less obvious to remove if you don't use it. I'd be supportive of this change if the template was updated to actually consume the environment variable as seen in the other templates (that call it `ENVIRONMENT`).

Thanks, updated!

samchungy (approved):

72636c (approved):

PR #1476, opened 2024-03-13
Remove `why-is-node-running`

72636c:
This is one idea I had; the other is to remove the dependency entirely.

https://seekchat.slack.com/archives/C03UM9GBGET/p1710315456217509


AaronMoat (approved):
I'm fine with either

PR #1480, opened 2024-03-18
RFC: Refresh ARM64 guide

72636c:
Follows on from #1350.

### DIFF
--- a/.changeset/fifty-toys-destroy.md
+++ b/.changeset/fifty-toys-destroy.md
@@ -10,7 +10,7 @@ Previously, the built-in templates made use of [`BUILDPLATFORM`](https://docs.do
 FROM --platform=${BUILDPLATFORM:-arm64} gcr.io/distroless/nodejs20-debian11
 ```
 
-1. Choose the platform of the host machine running the Docker build. A `vCurrent` Buildkite agent or Apple Silicon laptop will build under `arm64`, while an Intel laptop will build under `amd64`.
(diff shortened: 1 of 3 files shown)


.changeset/fifty-toys-destroy.md
```diff
@@ -10,7 +10,7 @@ Previously, the built-in templates made use of [`BUILDPLATFORM`](https://docs.do
 FROM --platform=${BUILDPLATFORM:-arm64} gcr.io/distroless/nodejs20-debian11
 ```
 
-1. Choose the platform of the host machine running the Docker build. A vCurrent SEEK build agent or Apple Silicon laptop will build under `arm64`, while an Intel laptop will build under `amd64`.
+1. Choose the platform of the host machine running the Docker build. A `vCurrent` Buildkite agent or Apple Silicon laptop will build under `arm64`, while an Intel laptop will build under `amd64`.
```
AaronMoat, line 13: Should we clarify this is just the default? x86 vCurrent is possible

AaronMoat (commented):
Skimmed only so warrants another set of eyes but looks great

.changeset/fifty-toys-destroy.md
72636c: Thanks for the catch, I initially missed that x86 vCurrent is the default and forgot to update this part.

etaoins (approved):
👏

PR #1481, opened 2024-03-19
Jest.mergePreset: Fix compatibility with `NodeNext`

72636c:
https://github.com/seek-oss/skuba/actions/runs/8338828529

Co-authored-by: Michael Taranto <mtaranto@seek.com.au>

### DIFF
--- a/jest/transform.js
+++ b/jest/transform.js
@@ -1,4 +1,5 @@
 const { defaults } = require('ts-jest/presets');
+const { ModuleResolutionKind } = require('typescript');
 
 const { tryParseTsConfig } = require('./tsConfig');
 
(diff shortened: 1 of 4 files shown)


jest/transform.js
```diff
@@ -38,10 +63,8 @@ module.exports.transform = Object.fromEntries(
             TS_JEST_PATH,
             {
               ...value[1],
-              isolatedModules:
-                value[1]?.isolatedModules ??
```
72636c, line 42: Removed some of this complexity because `defaults.transform` doesn't populate `isolatedModules` nor `tsconfig`. We can always update this in future if that changes.

samchungy (approved):

PR #1497, opened 2024-03-25
Migrate Skuba Dive

samchungy:


.npmrc
```diff
@@ -8,3 +8,4 @@ public-hoist-pattern[]="tsconfig-seek"
 # end managed by skuba
 
 public-hoist-pattern[]="remark-lint-*"
+ignore-workspace-cycles=true
```
samchungy, line 11: Silences this:

image

.prettierignore
```diff
@@ -8,4 +8,4 @@ pnpm-lock.yaml
 # end managed by skuba
 
 /integration/base/
-
+/packages/**/*/lib*
```
72636c, line 11: Is it worth putting this in `gitignore`? The `prettierignore` should then inherit it.

package.json
```diff
@@ -34,7 +34,8 @@
     "format": "pnpm --silent skuba format",
     "lint": "pnpm --silent skuba lint && pnpm --silent lint:md",
     "lint:md": "remark --frail --quiet .",
-    "release": "pnpm --silent build && changeset publish",
+    "prepack": "pnpm --silent build",
```
72636c, line 37: I thought that pnpm doesn't run `pre` scripts by default: https://pnpm.io/cli/run#differences-with-npm-run

We seem to be relying on these Changesets config options in our workflows:

https://github.com/seek-oss/skuba/blob/4049ce976c10b1002ace0339055684533c190882/.github/workflows/release.yml#L35-L38

https://github.com/seek-oss/skuba/blob/4049ce976c10b1002ace0339055684533c190882/.github/workflows/snapshot.yml#L31-L34

packages/eslint-config-skuba/package.json
```diff
@@ -31,6 +35,7 @@
   },
   "devDependencies": {
     "eslint": "^8.11.0",
+    "skuba": "workspace:*",
```
72636c, line 38: What happens if we don't have this here? (Or was this a merge conflict?)

package.json
samchungy: https://pnpm.io/cli/publish#life-cycle-scripts

Seems to run before these. I figure it will only run a build if it needs to based on what's changed?

packages/skuba-dive/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "lib/**/*.js.map",
+    "register.js"
+  ],
+  "scripts": {
+    "build": "pnpm skuba build",
+    "format": "pnpm skuba format",
+    "lint": "pnpm skuba lint",
+    "prepack": "pnpm skuba build",
```
samchungy, line 29: https://www.npmjs.com/package/skuba-dive/v/2.0.1-dive-migration-20240325115531?activeTab=code

Seems to work, unless the pre-publish script is the one running..?

packages/eslint-config-skuba/package.json
```diff
@@ -20,6 +20,11 @@
     "index.d.ts",
     "index.js"
   ],
+  "scripts": {
+    "format": "pnpm --silent skuba format",
+    "lint": "pnpm --silent skuba lint",
+    "skuba": "node --env-file=../../.env ../../lib/skuba"
```
samchungy, line 26: We could also call the build script every time, not sure what's better

.github/workflows/snapshot.yml
```diff
@@ -10,6 +10,7 @@ jobs:
     permissions:
       id-token: write
     runs-on: ubuntu-latest
+    timeout-minutes: 20
```
72636c, line 13: Ooh, worth adding to our template?

samchungy: Ooo probably, I accidentally hung the actions for a hour or 2 😅

packages/skuba-dive/package.json
samchungy: Confirmed this is required to populate a build artefacts for the repo

packages/skuba-dive/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "declaration": true,
+    "incremental": true,
+    "lib": ["ES2020"],
+    "module": "CommonJS",
+    "moduleResolution": "node",
+    "outDir": "lib",
+    "removeComments": false,
+    "target": "ES2020"
+  },
+  "exclude": ["lib*/**/*", "template/**/*"],
```
72636c, line 12: Not important, but I'm guessing we can do without this:

```suggestion
  "exclude": ["lib*/**/*"],
```

72636c (commented):
👏

Seems reasonable, just meta questions about how we want to structure monorepos so we can better support them going forward.

packages/skuba-dive/.dockerignore
```diff
@@ -0,0 +1,23 @@
...
+.DS_Store
+npm-debug.log
+yarn-error.log
+# end managed by skuba
+
+# Ignore .npmrc. This is no longer managed by skuba as pnpm projects use a managed .npmrc.
+# IMPORTANT: if migrating to pnpm, remove this line and add an .npmrc IN THE SAME COMMIT.
+# You can use `skuba format` to generate the file or otherwise commit an empty file.
+# Doing so will conflict with a local .npmrc and make it more difficult to unintentionally commit auth secrets.
+.npmrc
```
AaronMoat, line 23: should we "migrate" this to pnpm?

AaronMoat: actually can we delete the file? I'm not sure skuba would be managing it

AaronMoat: I confused myself, local `lint` doesn't run the subpackages. Feels weird 🤔

packages/skuba-dive/README.md
```diff
@@ -0,0 +1,101 @@
+# 🤿🌊
+
+[![npm package](https://img.shields.io/npm/v/skuba-dive)](https://www.npmjs.com/package/skuba-dive)
+[![Powered by skuba](https://img.shields.io/badge/🤿%20skuba-powered-009DC4)](https://github.com/seek-oss/skuba)
```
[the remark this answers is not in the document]
72636c:
```suggestion
[![Node.js version](https://img.shields.io/badge/node-%3E%3D%2014.18-brightgreen)](https://nodejs.org/en/)
```

72636c (approved):
Makes sense to merge this first, then make monorepo tweaks incrementally 👍

packages/skuba-dive/.dockerignore
samchungy: I'm thinking the changes to the subpackages won't be as frequent, so we could add a lint:all, or lint:packages for them? (Just thinking out loud)

PR #1508, opened 2024-03-26
WIP: Output .mjs for hybrid packages

72636c:
Snapshots via `cjs-mjs-output` tag: https://www.npmjs.com/package/skuba?activeTab=versions

```console
npm i skuba@cjs-mjs-output
```

---

This is an experimental feature that currently requires:

`package.json`

```diff
  {
    "files": [
      "lib*/**/*.d.ts",
      "lib*/**/*.js",
      "lib*/**/*.js.map",
+     "lib*/**/*.mjs",
+     "lib*/**/*.mjs.map"
    ],
-   "main": "./lib-commonjs/index.js",
+   "main": "./lib/index.js",
-   "module": "./lib-es2015/index.js",
+   "module": "./lib/index.mjs",
-   "types": "./lib-types/index.d.ts",
+   "types": "./lib/index.d.ts",
    "scripts": {
      "build": "skuba build-package"
    },
    "skuba": {
+     "build": "esbuild",
      "version": "x.x.x"
    },
  }
```

---

To do:

- [x] Consider forcing `declaration` and `removeComments` in this mode, in pursuit of zero config
- [ ] Is this our chance to get rid of `build-package` as a separate command? It may be better to consolidate on `build` with support for configuration; we can set `package.json#/skuba/build` to `esbuild` today, and we could add an option for e.g. `hybrid` output.
- [ ] How do we migrate existing packages off their tsc-based `lib-commonjs` and `lib-es2015` output? How safe will it be to switch to esbuild and modern targets?
- [ ] Should we investigate `skuba lint` autofixes or a `skuba migrate` for ^?


src/cli/build/esbuild.ts
```diff
@@ -46,20 +34,122 @@ export const esbuild = async (
...
+
+    case 'build-package':
+      await runBuild({
+        bundle: true,
+        compilerOptions: {
+          ...compilerOptions,
+          module: ModuleKind.ESNext,
+        },
+        debug,
+        entryPoints,
```
72636c, line 56: I think these `entryPoints` are not correct for `bundle: true`. Strictly speaking it should only output the root `lib/index.mjs` unless more `exports` are defined.

PR #1512, opened 2024-03-27
RFC: Mount `package.json` for Corepack

72636c:
Our Dockerfiles currently run `corepack enable pnpm` without any versioning information:

https://github.com/seek-oss/skuba/blob/bc2c69d3b85a6aba738c5d63d01f810a69b1a248/template/greeter/Dockerfile#L5

This may use a different version than is specified in package.json:

https://github.com/seek-oss/skuba/blob/bc2c69d3b85a6aba738c5d63d01f810a69b1a248/template/greeter/package.json

This can result in a subsequent `pnpm install --offline` to hang in CI on the following prompt:

```console
Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-8.15.4.tgz.

Do you want to continue? [Y/n]
```

PR #1515, opened 2024-04-02
Fix #1512

72636c:
I finally decided to read the Corepack documentation and skim its code. It turns out `corepack enable` doesn't actually do any installing.

https://github.com/nodejs/corepack?tab=readme-ov-file#utility-commands

### DIFF
--- a/.changeset/early-sloths-deliver.md
+++ b/.changeset/early-sloths-deliver.md
@@ -1,0 +1,46 @@
+---
+'skuba': patch
+---
+
+template: Install specific pnpm version via Corepack
+
+Previously, our Dockerfiles ran `corepack enable pnpm` without installing a specific version. This does not guarantee installation of the pnpm version specified in `package.json`, which could cause a subsequent `pnpm install --offline` to run Corepack online or otherwise hang on stdin:
+
+```dockerfile
+FROM --platform=arm64 node:20-alpine
+
+RUN corepack enable pnpm
+```
+
+```json
+{
+  "packageManager": "pnpm@8.15.4",
+  "engines": {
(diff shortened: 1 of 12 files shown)


.changeset/early-sloths-deliver.md
```diff
@@ -0,0 +1,46 @@
...
+
+To avoid this issue, modify (1) Buildkite pipelines to cache on `package.json`, and (2) Dockerfiles to mount `package.json` and run `corepack install`:
+
+```diff
+  seek-oss/docker-ecr-cache#v2.1.0:
+    cache-on:
+     - .npmrc
++    - package.json
+     - pnpm-lock.yaml
+```
```
72636c, line 38: @samchungy I won't release this until we sort out seek-oss/docker-ecr-cache-buildkite-plugin#51, so you can modify this changeset later.

Alternatively, we could actually revert #1512. It _feels_ wrong but `pnpm install --offline` can invoke the Corepack shim and install the appropriate version as required. The reason one of my repos hit a stdin hang is because it wasn't using `propagate-environment` so Corepack thought it wasn't running in `CI=true`.

AaronMoat: I defer to sam

samchungy: Nah I think what you've got here makes more sense. If we're pinning the version of pnpm, we want to ensure it is using the pinned version, just in case a bad release creeps in.

samchungy: I can probably write up a patch to target this

AaronMoat (approved):

PR #1534, opened 2024-04-10
Patch Pipelines and Dockerfiles

samchungy:
Patches the [issue](https://github.com/seek-oss/skuba/pull/1515) in our consumer repos who have already migrated to pnpm

Tested in one of my repos so far:

image


AaronMoat (approved):
👏

PR #1541, opened 2024-04-16
Replace CDK context config

samchungy:
Thoughts?

This is pretty clunky and involves defining config awkwardly twice.

One time in untyped JSON and then again with zod.

This cuts out both and just relies on standard TypeScript


template/lambda-sqs-worker-cdk/infra/appStack.test.ts
```diff
@@ -31,14 +15,22 @@ jest.useFakeTimers({
   now: new Date(currentDate),
 });
 
-it.each(contexts)(
+afterEach(() => {
+  jest.resetModules();
+});
+
+it.each(['dev', 'prod'])(
   'returns expected CloudFormation stack for $stage',
```
72636c:
```suggestion
  'returns expected CloudFormation stack for %s',
```

72636c (approved):

Issue #1093, opened 2023-01-26
Recommended Extensions for Skuba Projects


samchungy:
We can create a default `.vscode` folder with an `extensions.json` file:

```json
{
    "recommendations": [
        "esbenp.prettier-vscode",
        "dbaeumer.vscode-eslint"
    ]
}
```

to recommend plugins we suggest. It comes up with a prompt when you open the project which I thought is kinda neat.

mrjono1:
example with more recommendations https://github.com/SEEK-Jobs/noti-experience-resources/blob/master/.vscode/settings.json#L29

PR #1547, opened 2024-04-23
Add extension recommendations

GeordieEK:
Addresses issue #1093 

`skuba init` now creates a .vscode folder and inserts extension recommendations. Implementation written here will overwrite recommendations if they exist already however as it's part of init I think that's ok.

I have opted for a patch changeset as per contribution guidelines of minor change involving `init`.


samchungy:
Hey! Thanks for the PR! I love the attempt

I think a slightly simpler option may be just to create the files under our `base` folder section. I think that should work.

eg.
https://github.com/seek-oss/skuba/blob/main/template/base/.github/CODEOWNERS

GeordieEK:
Thanks! 

That's definitely a simpler idea.

The issue I'm running into at the moment is the inclusion filter is excluding `.vscode/*` as it's in the .gitignore

Some possible solutions I've identified:
- Don't use the .gitignore in the filter but filter based on an exclusionFilter file or similar for more granular control
    - Less code change but feels more hacky 
- Modify the inclusion filter function itself to allow exceptions (i.e files in gitignore that we do actually want to copy)
    - This is potentially too big of a change for a small feature, but may come in handy later? 
    - I also don't think this would classify as a patch anymore

I have done the latter in commit 97b487c87b7691778a8519d42490aaab964b19c4

AaronMoat:
Hey @GeordieEK - I'm liking the direction! 

I wonder if rather than modifying the code - could we get away with explicitly allowing this file? 

For example, adding `!.vscode/extensions.json` after the `.vscode/ignore*` in the _.gitignore template?

AaronMoat:
I think so - struggling to see a huge downside, though open to opinions of course

GeordieEK:
Yeah I think it's unlikely anyone will really be bothered by that inclusion.

I would say the level of change to the codebase fits the utility of the improvement better. 

Changing a utility function for this small of a benefit is a bit extreme... fun first issue though!

samchungy:
I think including extensions.json in the git makes sense. The main reason the vscode folder is ignored is mostly to stop user settings from constantly being committed, however, this isn't really one of those files

GeordieEK:
Simplified implementation implemented at #1556

Issue #1401, opened 2024-01-23
Use `VERSION` in lambda-sqs-worker-cdk template


samchungy:
Ah yes, I think that's right re: `VERSION`. Maybe we should extend the CDK template to use the version variable too

CDK

https://github.com/seek-oss/skuba/blob/454710c4ab068c073918ec2426896756169ce22d/template/lambda-sqs-worker-cdk/src/app.ts#L4-L6

Serverless

https://github.com/seek-oss/skuba/blob/454710c4ab068c073918ec2426896756169ce22d/template/lambda-sqs-worker/src/config.ts#L48

https://github.com/seek-oss/skuba/blob/454710c4ab068c073918ec2426896756169ce22d/template/lambda-sqs-worker/src/framework/logging.ts#L14-L17

_Originally posted by @72636c in https://github.com/seek-oss/skuba/pull/1392#discussion_r1462597460_

PR #1548, opened 2024-04-23
Add version variable to cdk template logger

GeordieEK:
Addresses issue #1401

Extend CDK template to use VERSION variable in logger.

I also re-organised cdk config slightly so that it better matches serverless config at a glance.

I have opted for a patch changeset as per contribution guidelines of minor changes involving templates.


template/lambda-sqs-worker-cdk/src/app.ts
```diff
@@ -1,8 +1,10 @@
 import createLogger from '@seek/logger';
 import type { SQSEvent, SQSHandler } from 'aws-lambda';
+import { config } from 'infra/config';
```
72636c, line 3: Thanks for the contribution @GeordieEK!

@samchungy I know you have opinions on how config is structured for CDK applications. Should we share config between the infra stack & runtime app?

samchungy: Hmmm I'm trying to think if there's any issues with them being separate. It does feel wrong to intermix them somehow.

Should they be separate? 🤔

samchungy: I think we want to keep them separate. If there are some environment variables we want to set just for the CDK infra deployment, we would always need to set them even if we were simply running the application.

I think we should keep our app config separate from our infra config. Especially given the uptake in monorepos

GeordieEK: I've separated infra and runtime configs but can't figure out how to fix linting errors

samchungy: Ah good attempt! The template can be a little tough to work with tbh. I'll have a quick looksie tomorrow morning

samchungy: I tend to run the test:template command to essentially scaffold the template and then edit it within there

GeordieEK: Yeah! I tried the test:template command, very nice. It makes clear what is causing it but I'm not sure why it's complaining for the cdk template but not the others that appear to me to have similar structure. Appreciate it :)

samchungy: I've pushed up some changes. If you jump to ../tmp-skuba/tmp-lambda-sqs-worker-cdk you should be able to see linting errors directly within your IDE.

.changeset/slow-seahorses-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+template/lambda-sqs-worker-cdk: Add VERSION variable to `lambda-sqs-worker-cdk` logger
```
72636c, line 5:
```suggestion
template/lambda-sqs-worker-cdk: Add `version` property to logs
```

template/lambda-sqs-worker-cdk/src/app.ts
```diff
@@ -1,8 +1,20 @@
...
+    environment: config.environment,
+    version: config.version,
+  },
+
+  level: config.logLevel,
+
+  name: config.name,
+
+  transport:
+    config.environment === 'local' ? { target: 'pino-pretty' } : undefined,
```
72636c, line 17: Do we need to declare `pino-pretty` as a dev dep?

72636c (approved):

template/lambda-sqs-worker-cdk/src/app.ts
GeordieEK: Ahh, that's the trick I was missing. Thanks so much!

template/lambda-sqs-worker-cdk/src/config.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  [dev]: () => ({
+    ...configs[prod](),
+
+    logLevel: 'debug',
+  }),
+
+  [prod]: () => ({
+    logLevel: 'info',
+    name: Env.string('SERVICE'),
```
samchungy, line 43: We also need to also declare these in `infra/config.ts`

You can mostly copy paste the `skuba-dive` setup over.

template/lambda-sqs-worker-cdk/infra/config.ts
```diff
@@ -33,9 +35,10 @@ export const configs: Record<Environment, Config> = {
       environment: {
         ENVIRONMENT: 'prod',
       },
+      SERVICE: '<%- serviceName %>',
```
samchungy, line 38: This will have to be in environment to be passed to the worker.

Could we also add VERSION but we're going to take in an environment variable which will come from our CI

https://github.com/seek-oss/skuba/tree/main/packages/skuba-dive#env

Could we also add this env var into the pipeline
https://github.com/seek-oss/skuba/blob/81780d930e8c2dd3f720f52223895580407af61d/template/lambda-sqs-worker/.buildkite/pipeline.yml#L50

samchungy: This is looking excellent

GeordieEK: Ah yep! I think I get you now., I think I'd lost sight of the original purpose of the PR and was trying to avoid editing it too much... I hope this is better.

I don't think I can run the snapshot actions but I updated the relevant snapshot manually

template/lambda-sqs-worker-cdk/infra/config.ts
```diff
@@ -22,6 +26,8 @@ export const configs: Record<Environment, Config> = {
       reservedConcurrency: 2,
       environment: {
         ENVIRONMENT: 'dev',
+        SERVICE: Env.string('SERVICE', { default: '<%- serviceName %>' }),
```
samchungy, line 29:
```suggestion
        SERVICE: <%- serviceName %>,
```

template/lambda-sqs-worker-cdk/infra/config.ts
```diff
@@ -32,10 +38,12 @@ export const configs: Record<Environment, Config> = {
       reservedConcurrency: 20,
       environment: {
         ENVIRONMENT: 'prod',
+        SERVICE: Env.string('SERVICE', { default: '<%- serviceName %>' }),
```
samchungy, line 41:
```suggestion
        SERVICE: <%- serviceName %>,
```

samchungy (approved):

samchungy:
Thanks for the hard work 🙇

AaronMoat (approved):

PR #1552, opened 2024-04-25
Fix package manager detection

AaronMoat:
In https://github.com/seek-oss/skuba/pull/1550, I found I had test snapshot diffs locally and couldn't work it out. I suspect there's funky things going on with detect-package-manager, like checking for global installs, that are somewhat unnecessary for our use cases given the strong prevalence of lockfiles with skuba projects.


packages/skuba-dive/.npmrc
samchungy, line 1: These might break changesets

AaronMoat: huh why?

AaronMoat: Do you mean in consuming repos or in skuba itself?

samchungy (approved):
In principle lgtm

packages/skuba-dive/.npmrc
AaronMoat: they did indeed break changesets, was able to get a snapshot running on a branch off this (with a dive changeset) https://github.com/seek-oss/skuba/actions/runs/8892072630

AaronMoat:
Snapshot is happy - are you able to take another quick look @samchungy? https://github.com/seek-oss/skuba/actions/runs/8964626049

samchungy (approved):

PR #1554, opened 2024-05-05
Respect .gitignore when trying to generate files

AaronMoat:


package.json
```diff
@@ -36,11 +36,11 @@
     "lint:md": "remark --frail --quiet .",
     "release": "pnpm --silent build && changeset publish",
     "skuba": "pnpm --silent build && pnpm --silent skuba:exec",
-    "skuba:exec": "node --env-file=.env lib/skuba",
+    "skuba:exec": "node --experimental-vm-modules lib/skuba",
```
AaronMoat, line 39: Using the .env file was breaking when using a VSCode debug terminal, because that sends NODE_OPTIONS. I can't see any downsides to inlining it?

package.json
```diff
@@ -36,11 +36,11 @@
...
     "stage": "changeset version && node ./.changeset/inject.js && pnpm format",
-    "test": "pnpm --silent skuba test --selectProjects unit",
+    "test": "pnpm --silent skuba test --selectProjects unit --",
```
AaronMoat, line 41: Allows you to do `pnpm test <file>`, same as below. The problem is that --selectProjects accepts varargs

src/cli/lint.int.test.ts
```diff
@@ -70,6 +70,7 @@ const stdout = (randomMatcher: RegExp) => {
 const prepareTempDirectory = async (baseDir: string, tempDir: string) => {
   await copy(baseDir, tempDir);
   process.chdir(tempDir);
+  await git.init({ fs, dir: tempDir });
```
AaronMoat, line 73: To get it to not pay attention to skuba's root gitignore

src/api/git/index.ts
```diff
@@ -9,3 +9,4 @@ export { getOwnerAndRepo } from './remote';
 export { push } from './push';
 export { fastForwardBranch } from './pull';
 export { reset } from './reset';
+export { isFileGitIgnored } from './isFileGitIgnored';
```
samchungy, line 12: Could we add a changeset for the extra git api added?

samchungy (approved):

PR #1560, opened 2024-05-19
template: Make all configuration values explicit

72636c:
Follows on from seek-oss/datadog-custom-metrics#286.


AaronMoat (approved):
🙇 still DRY (do repeat yourself)

PR #1561, opened 2024-05-22
Secret helpers

jgrigg:
This is an attempt to plug a gap we have in Automat right now around injecting secrets as environment variables.

This aligns with [RFC007](https://backstage.myseek.xyz/docs/default/component/rfcs/RFC007-Secrets-Management-Securing-Sensitive-Secrets/#preferred-solution-aws-secrets-manager) where AWS Secrets Manager is our preferred solution at SEEK.

More info in this thread https://seekchat.slack.com/archives/C037V5FSD18/p1716254655390659?thread_ts=1716253996.973889&cid=C037V5FSD18


packages/skuba-dive/package.json
```diff
@@ -32,10 +32,12 @@
     "test:ci": "pnpm skuba test --coverage"
   },
   "dependencies": {
+    "@aws-sdk/client-secrets-manager": "^3.580.0",
```
jgrigg, line 35: I suspect this is undesirable. could be a peer dep instead maybe.

packages/skuba-dive/README.md
```diff
@@ -64,6 +65,25 @@ export const flag = Env.boolean('FLAG');
...
+import { Secret } from 'skuba-dive';
+
+export const apiKey = Secret.string('api-key', { default: 'local' });
+// string | 'local'
+
+export const signature = Secret.binary('signature', { default: undefined });
+// Uint8Array | undefined
+```
+
+Each function will throw if the secret cannot be fetched or is messing a value when `opts.default` is not provided.
```
gitlon, line 85:
```suggestion
Each function will throw if the secret cannot be fetched or is missing a value when `opts.default` is not provided.
```

just a typo in the README change

PR #1578, opened 2024-06-08
Add Templates as Workspaces

samchungy:
Gives us a better experience while working with them

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "npm-run-path": "^4.0.1",
     "npm-which": "^3.0.1",
     "picomatch": "^4.0.0",
-    "prettier": "~3.2.5",
+    "prettier": "~3.3.0",
     "prettier-plugin-packagejson": "^2.4.10",
     "read-pkg-up": "^7.0.1",
     "semantic-release": "^22.0.12",
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "format": "pnpm --silent skuba format",
     "lint": "pnpm --silent skuba lint && pnpm --silent lint:md",
     "lint:md": "remark --frail --quiet .",
+    "prepack": "pnpm --filter './template/**' exec rm -rf node_modules",
     "release": "pnpm --silent build && changeset publish",
     "skuba": "pnpm --silent build && pnpm --silent skuba:exec",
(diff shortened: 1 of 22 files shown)


template/koa-rest-api/package.json
```diff
@@ -21,11 +22,10 @@
     "@opentelemetry/propagator-b3": "^1.25.0",
     "@opentelemetry/sdk-node": "^0.52.0",
     "@seek/logger": "^6.0.0",
-    "aws-sdk": "^2.1039.0",
```
samchungy, line 24: Not used? otel seems to support v3.

template/koa-rest-api/package.json
```diff
@@ -21,11 +22,10 @@
...
     "hot-shots": "^10.0.0",
     "koa": "^2.13.4",
     "koa-bodyparser": "^4.3.0",
-    "koa-compose": "^4.2.0",
```
samchungy, line 28: image
image

pnpm-lock.yaml
72636c, line 1: One thing with the old model is that our CI broke when a clean `init` without a lockfile broke.

Is it worth scripting up frequent, automerged lock file maintenance so we get a similar feedback loop?

template/express-rest-api/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "<%- serviceName %>",
```
72636c, line 2: Should we put a `@seek/` scope out of paranoia?

I've seen us pay out bounties for "dependency confusion" even though I tend to disagree with the premise.

72636c (commented):
👏

72636c (approved):

pnpm-lock.yaml
samchungy: image

It always installs the latest in the integration test for the templates so we're safe there! I tested a downgrade of zod

package.json
```diff
@@ -34,6 +34,7 @@
     "format": "pnpm --silent skuba format",
     "lint": "pnpm --silent skuba lint && pnpm --silent lint:md",
     "lint:md": "remark --frail --quiet .",
+    "prepack": "pnpm --filter './template/**' exec rm -rf node_modules",
```
samchungy, line 37: https://github.com/pnpm/pnpm/pull/7301

Workaround hack for this

PR #1582, opened 2024-06-13
Remove duplicate check runs

samchungy:
Disables run of checks on successive pushes to the pull request

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -2,7 +2,7 @@ name: Validate
 
 on:
   pull_request:
-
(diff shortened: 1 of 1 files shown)


.github/workflows/validate.yml
72636c, line 1: Does this allow us to build successive commits on forks?

72636c (approved):

.github/workflows/validate.yml
samchungy: Think so, I'll give it a shot tomorrow

PR #1586, opened 2024-06-14
Skip Excessive Actions Runs

samchungy:
Disable extra runs from occurring whenever we push to the repo.

**On a fork:**

This works for Pull Requests raised and pushes to the fork

image

**On a branch within the repo**

image

### DIFF
--- a/.changeset/hip-mice-punch.md
+++ b/.changeset/hip-mice-punch.md
@@ -1,0 +1,5 @@
+---
+'skuba': patch
+---
+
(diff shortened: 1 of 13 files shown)


72636c (approved):
🥲 thanks!

Should we update the template too: https://github.com/seek-oss/skuba/blob/528f58fb58294bf4b1dc594f025181f53b1a252c/template/oss-npm-package/.github/workflows/validate.yml#L1-L16

Issue #1563, opened 2024-06-01
Setting "module": "node16" causes tsc to complain


samchungy:
<img width="1343" alt="image" src="https://github.com/seek-oss/skuba/assets/18017094/8a8d5309-f356-40ba-a255-86f165cabb89">

My head is brokey

https://x.com/atcb/status/1634653541607563271?t=JrCYikngA3oq5CfMb_0iXQ&s=19

PR #1589, opened 2024-06-17
Explicitly declare return types

samchungy:
Resolves https://github.com/seek-oss/skuba/issues/1563


AaronMoat (approved):
I don't really understand the need for this but seems fine

samchungy:
> I don't really understand the need for this but seems fine

something something esm/cjs cranky

samchungy:
> I don't really understand the need for this but seems fine

We ultimately should be exporting different types for cjs, and esm:

image

But because we're only shipping one, the TypeScript compiler is going why are you trying to import in my cjs world.

72636c (approved):

PR #1597, opened 2024-06-25
Fix [object Object] output

AaronMoat:


samchungy (approved):

.changeset/funny-sloths-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+Fix [object Object] output appearing during `skuba configure`
```
samchungy, line 5:
```suggestion
configure: Fix [object Object] output appearing during `skuba configure`
```

AaronMoat: I will never remember this

PR #1601, opened 2024-07-02
Fix: enable secure header middleware by default

tadhglewis:
Idk why we _wouldn't_ have this? 

> 
> As such, it's important that this middleware is only used under these circumstances:
> 
> 1. The app's domain does not mix HTTP and HTTPS.
> 
> 2. The response is an API response, i.e. it is not being directly rendered by the browser.
> 
> 3. The responses have an accurate `Content-Type` header.

https://github.com/seek-oss/koala/tree/master/src/secureHeaders


tadhglewis:
Related: https://github.com/seek-oss/skuba/issues/1602

AaronMoat:
Hoping for other opinions, but I don't mind the change. I'd like your thoughts on whether adding some form of warning / documentation linking is suitable here?

template/koa-rest-api/src/framework/server.ts
72636c, line 1: @etaoins do you have any thoughts as Koala BDFL?

etaoins: I don't mind this either. This is the big one because it can break loading all HTTP resources hosted on the same domain:
> The app's domain does not mix HTTP and HTTPS

However, that's become even less common in the 5 years since this documentation was written. The other caveats only have a local effect and would presumably be caught during testing.

etaoins (approved):

72636c (approved):

template/koa-rest-api/src/framework/server.ts
```diff
@@ -39,10 +39,7 @@ export const createApp = <State, Context>(
   ...middleware: Array<Koa.Middleware<State, Context>>
 ) =>
   new Koa()
-    // TODO: consider using a middleware that adds secure HTTP headers.
-    // https://github.com/seek-oss/koala/tree/master/src/secureHeaders
-    // https://github.com/venables/koa-helmet
-    // .use(SecureHeaders.middleware)
+    .use(SecureHeaders.middleware)
```
AaronMoat, line 42:
```suggestion
    // Read: https://github.com/seek-oss/koala/tree/master/src/secureHeaders
    .use(SecureHeaders.middleware)
```

AaronMoat (approved):
Can you add a patch changeset, please?

.changeset/heavy-scissors-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+Enable secure headers middleware by default
```
AaronMoat, line 5:
```suggestion
template/koa-rest-api: Enable secure headers middleware by default
```

AaronMoat (approved):

PR #1603, opened 2024-07-02
template/koa-rest-api: src/framework/middleware => src/framework/bodyParser

72636c:
This "middleware" file is a bit weird because we also have middleware in other files like `src/framework/logging` and `src/framework/server`.

I don't think this is worth a changeset.


AaronMoat (approved):
Should we swap out the package while we're here, or stick to the popular one? https://github.com/koajs/bodyparser/issues/161

PR #1604, opened 2024-07-02
RFC: Do not modify config files if version is outdated

72636c:
I noticed some recent commits where config file changes were going back and forth. However, after writing this I'm not sure that it is actually worth it, given e.g. a Prettier change between versions would similarly cause such churn if local dependencies are outdated. Thoughts?

### DIFF
--- a/src/cli/lint/internalLints/refreshConfigFiles.test.ts
+++ b/src/cli/lint/internalLints/refreshConfigFiles.test.ts
@@ -5,6 +5,7 @@ import * as fsExtra from 'fs-extra';
 import { Git } from '../../..';
 import { log } from '../../../utils/logging';
 import { detectPackageManager } from '../../../utils/packageManager';
+import * as packageAnalysis from '../../configure/analysis/package';
 import * as project from '../../configure/analysis/project';
 
 import {
@@ -48,7 +49,17 @@ const givenMockPackageManager = (command: 'pnpm' | 'yarn') => {
 
 jest.mock('../../../utils/packageManager');
 
+let getDestinationManifest = jest.spyOn(
+  packageAnalysis,
+  'getDestinationManifest',
+);
+
 beforeEach(() => {
+  getDestinationManifest = jest.spyOn(
+    packageAnalysis,
+    'getDestinationManifest',
+  );
+
   jest
     .spyOn(console, 'log')
     .mockImplementation((...args) => stdoutMock(`${args.join(' ')}\n`));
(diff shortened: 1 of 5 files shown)


src/cli/lint/internalLints/refreshConfigFiles.test.ts
```diff
@@ -252,6 +288,27 @@ The .dockerignore file is out of date. Run \`pnpm exec skuba format\` to update
...
+
+      await expect(refreshConfigFiles('format', log)).resolves.toEqual({
+        ok: true,
+        fixable: false,
+      });
+
+      expect(stdout()).toBe('');
+
+      expect(writeFile).not.toHaveBeenCalled();
+    });
```
tadhglewis, line 310: Is it possible for us to tell the user to do a `ni`? Having out of date local packages can cause other issues

tadhglewis:
Thx for the QoL fix! It's definitely not ideal if skuba formats in CI but gets overwritten by the dev again

AaronMoat:
Wonder if global skuba should delegate to local skuba if it can find it 🤔

tadhglewis:
@AaronMoat 

Scenario

1. Dev pulls master and does `pnpm format`
2. Dev pushes
3. Skuba formats due to dev format using out of date skuba
4. Dev pulls and then re-pushes his bad format

Repeat

src/cli/lint/internalLints/refreshConfigFiles.ts
```diff
@@ -74,12 +76,28 @@ export const REFRESHABLE_CONFIG_FILES: RefreshableConfigFile[] = [
...
+  const [currentVersion, manifest, gitRoot] = await Promise.all([
+    getSkubaVersion(),
     getDestinationManifest(),
     Git.findRoot({ dir: process.cwd() }),
   ]);
 
+  const manifestVersion = manifestSkubaVersion(manifest);
+
+  process.stdout.write(
+    `${JSON.stringify({ currentVersion, manifestVersion })}\n`,
```
AaronMoat, line 89: Should this have words on it?

AaronMoat (approved):

src/cli/lint/internalLints/refreshConfigFiles.ts
72636c: WTF

AaronMoat:
I actually think it could warrant a log 🤷

PR #1608, opened 2024-07-08
Fix pnpm guide

samchungy:
Turns out this is required. Oops. We removed it in another repo due to a different change


AaronMoat (approved):
is this for the nested package jsons? Should we be mounting them individually not the whole contents?

samchungy:
> is this for the nested package jsons? Should we be mounting them individually not the whole contents?

It's actually everything right now. We need the built contents of the package, so at the moment we're just copying the whole dir including TypeScript files right now. I want to move to copying only a packed version of the package into the deployment but that will be post moving to pnpm.

PR #1621, opened 2024-07-29
Fix pnpm guide

AaronMoat:
We need to run format twice now, once to ensure .npmrc is moved to the bottom of the file if not already, and then a second time to actually synthesise the file now it's now gitignored, after https://github.com/seek-oss/skuba/pull/1554


lantrix (approved):
Thanks for that

samchungy (approved):

PR #1623, opened 2024-07-30
Replace ts-node with tsx

AaronMoat:
~Completely untested but opening up to early feedback on the principle.~

Somewhat tested now, are we happy with major+upgrade at own risk laziness, or are there things we can do to get more confidence?


src/cli/configure/dependencies/skubaDeps.ts
```diff
@@ -5,21 +5,22 @@ const DEV_DEPENDENCIES = [
   '@seek/seek-module-toolkit',
   'eslint-config-seek',
   'nodemon',
-  'tslint',
+  'ts-node-dev',
```
samchungy, line 8: do we remove this now?

samchungy: Oh nvm

samchungy (approved):
Looks aight from a quick look

src/cli/node.ts
```diff
@@ -41,13 +36,6 @@ export const node = async () => {
     );
   }
 
-  // REPL with `ts-node` to support import statements.
-  return tsNode
-    .createRepl({
-      service: tsNode.register({
-        require: ['dotenv/config', 'tsconfig-paths/register'],
```
72636c, line 48: Is the thinking that these don't matter for a REPL, or is the functionality baked in to tsx? (if we're keeping a REPL around, being able to `import 'src/api/blah'` could still be nice...)

72636c (approved):

src/cli/node.ts
AaronMoat: There is no thinking 😂 I need to test this still. Good question though

.changeset/empty-cherries-admire.md
```diff
@@ -0,0 +1,14 @@
+---
+'skuba': major
+---
+
```
AaronMoat, line 4: @72636c would you mind wordsmithing this? I'll proceed to merge without significant manual testing in SEEK repos; I hope to be able to get out https://github.com/seek-oss/skuba/pull/1537 as a second major changeset and bundle them in a release, so testing in real repos is likely to be valuable as a combined exercise

Issue #1493, opened 2024-03-25
Renovate + changesets investigation


AaronMoat:
Look into: 

- Auto generating changesets for renovates in the skuba repository, so we can't forget to generate one 
- Using `gitIgnoredAuthors` for the above to not lock the PRs after first commit.
- Extending `gitIgnoredAuthors` for the general SEEK ecosystem in combination with skuba autofixes, to not lock renovate PRs after first commit

AaronMoat:
We can try again when v38 is released: https://github.com/renovatebot/renovate/issues/27992

samchungy:
Can we delete https://github.com/seek-oss/skuba/blob/3310e6ed714cd4b06b03890c5c900e33b20b9ab9/src/cli/lint/autofix.ts#L74 if it lands?

AaronMoat:
Ah nice. Yes, likely, though it means that we're relying on users using rynovate* for it to work nicely. Which, maybe is ok

(*unless this makes momentum: https://github.com/seek-oss/renovate-config-seek/pull/21)

PR #1633, opened 2024-08-11
Try to generate changesets on Renovate PRs

AaronMoat:
This is difficult to test without merging 🙈 

Last bit to resolve https://github.com/seek-oss/skuba/issues/1493


.github/workflows/renovate-changeset.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Configure Git
+        run: |
+          git config --global user.email 34733141+seek-oss-ci@users.noreply.github.com
+          git config --global user.name 'Github changeset workflow'
+      - name: Generate changeset
+        uses: actions/github-script@v7
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        with:
+          script: |
```
zbrydon, line 30: Do actions requite this to be inline?

zbrydon (approved):

.github/workflows/renovate-changeset.yml
AaronMoat: Not directly, but it's a bit gross either way https://github.com/actions/github-script?tab=readme-ov-file#run-a-separate-file

.github/workflows/renovate-changeset.yml
```diff
@@ -0,0 +1,87 @@
...
+
+            const { stdout: shortHash } = await exec.getExecOutput(
+              'git rev-parse --short HEAD',
+            );
+            const fileName = `.changeset/renovate-${shortHash.trim()}.md`;
+
+            await createChangeset(fileName, prTitle, packageNames);
+            await exec.exec('git', ['add', fileName]);
+            await exec.exec('git commit -C HEAD --amend --no-edit');
+            await exec.exec('git push --force');
```
samchungy, line 87: Not sure if we're requiring commit signing here but this doesn't look like it'll sign yeah?

AaronMoat: TBH unsure - as it's replacing the commit. I don't think we require it atm

AaronMoat:
Going to try and likely revert 😆

Issue #1553, opened 2024-05-01
Docker Compose `version` is obsolete 


samchungy:
We should remove them from our templates

<img width="1084" alt="image" src="https://github.com/seek-oss/skuba/assets/18017094/52301592-b764-46cb-8b66-bee8e0e38b2c">

https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element-obsolete

Could be worth writing an autofix

PR #1638, opened 2024-08-20
Remove docker-compose version autofix

zbrydon:
https://github.com/seek-oss/skuba/issues/1553
> Could be worth writing an autofix

🙇

### DIFF
--- a/src/cli/lint/internalLints/upgrade/patches/8.2.1/patchDockerCompose.ts
+++ b/src/cli/lint/internalLints/upgrade/patches/8.2.1/patchDockerCompose.ts
@@ -37,7 +37,7 @@ const patchDockerComposeFiles: PatchFunction = async ({
   const dockerComposeFiles = await fetchFiles(maybeDockerComposeFiles);
 
   const dockerComposeFilesToPatch = dockerComposeFiles.filter(({ contents }) =>
-    contents.includes('version:'),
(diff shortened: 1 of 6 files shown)


src/cli/lint/internalLints/upgrade/patches/8.2.1/patchDockerCompose.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (!maybeDockerComposeFiles.length) {
+    return {
+      result: 'skip',
+      reason: 'no docker-compose files found',
+    };
+  }
+
+  const dockerComposeFiles = await fetchFiles(maybeDockerComposeFiles);
+
+  const dockerComposeFilesToPatch = dockerComposeFiles.filter(({ contents }) =>
```
AaronMoat, line 39: I think if we wanted to be rigorous about this, we'd do the .replace up here (or .match with the regex), and then check if any are different; technically you could have version: but not match the regex below.

AaronMoat (approved):
This looks good! If you can be bothered, it might be worth adding a couple more tests for 'lint' mode as well as rarer cases where we SHOULDN'T replace, like if there's a version: environment variable elsewhere in the file (indented)

AaronMoat (approved):

Issue #996, opened 2022-10-11
Support esbuild bundling


72636c:
https://github.com/seek-oss/skuba/blob/5097cd548a89fdad401937bd4903ff41b7557bfc/src/cli/build/esbuild.ts#L92-L96

72636c:
https://github.com/pinojs/pino/blob/main/docs/bundling.md#esbuild-plugin

PR #1644, opened 2024-08-31
Use aws-codedeploy-hook. Remove preHook.ts and postHook.ts

### LINKED ISSUE #1640: lambda-sqs-worker-cdk: use aws-codedeploy-hooks
Would be good to avoid copy-paste (via template) with https://github.com/seek-oss/aws-codedeploy-hooks.

aurelius0523:
## Overview
Use `aws-codeploy-hooks` and remove existing `codedeploy` code. Resolves #1640 

### Note 

I am unsure if this approach should be used to replace current implementation in `skuba`. `skuba` currently creates `pre` and `post` hooks that are not account-global which would allow the template consumer to do more specific smoke tests like checking for existence of a `dynamodb` table. 

By using `aws-codedeploy-hooks` as it is now we lose that ability. Happy to iterate `aws-codedeploy-hooks` to support that or update this PR if I am just missing how to do that


template/lambda-sqs-worker-cdk/infra/appStack.ts
```diff
@@ -74,6 +72,7 @@ export class AppStack extends Stack {
       target: 'node20',
       // aws-sdk-v3 is set as an external module by default, but we want it to be bundled with the function
       externalModules: [],
+      nodeModules: ['@seek/logger', 'skuba-dive'],
```
aurelius0523, line 75: Added these because I got the following error and I am not sure what is causing it when i run `pnpm run deploy`. Happy to fix this if anyone happens to know why

```
[ERROR] Could not resolve "@seek/logger"

    src/app.ts:1:25:
      1 │ import createLogger from '@seek/logger';
        ╵                          ~~~~~~~~~~~~~~

  The Yarn Plug'n'Play manifest forbids importing "@seek/logger" here because it's not listed as a
  dependency of this package:
```

AaronMoat:
```suggestion
```

Based on the error, I suspect you created a project with yarn >v1? I haven't seen this error but have used the template and haven't found this problem. Could you try with yarn 1 or pnpm?

template/lambda-sqs-worker-cdk/infra/index.ts
```diff
@@ -15,3 +16,11 @@ new AppStack(app, 'appStack', {
     // 'seek:system:name': 'TODO: add system name',
   },
 });
+
+// eslint-disable-next-line no-new
+const hookStack = new HookStack(app, 'hookStack');
```
AaronMoat, line 21: This is an interesting one; if we're recommending deploying per-application, it probably needs a unique stack name / function names / etc baked in. I'm unsure on the best approach here!

AaronMoat (commented):
Thanks for tackling this one! 

Any chance you could add a changeset?

> which would allow the template consumer to do more specific smoke tests like checking for existence of a dynamodb table.

To your note on this, the idea of these pre-hooks is that they don't contain the logic themselves, but instead ask the lambda function being deployed to check. If you wanted to ensure dynamodb existence, that could be part of the worker's smokeTest function.

template/lambda-sqs-worker-cdk/infra/appStack.ts
samchungy: Might be related? https://github.com/seek-oss/skuba/issues/996#issuecomment-2319598106

template/lambda-sqs-worker-cdk/package.json
```diff
@@ -16,6 +16,8 @@
     "@aws-sdk/client-codedeploy": "^3.363.0",
     "@aws-sdk/client-lambda": "^3.363.0",
     "@aws-sdk/client-sns": "^3.363.0",
+    "@aws-sdk/client-sso-oidc": "^3.637.0",
```
samchungy, line 19: question: do we need this?

template/lambda-sqs-worker-cdk/infra/appStack.ts
aurelius0523: Thanks for that. Tried with `pnpm` and `yarn 1` and all looks good

template/lambda-sqs-worker-cdk/package.json
aurelius0523: We do not. Removed. 🙏

template/lambda-sqs-worker-cdk/infra/index.ts
samchungy: There's normally a `from` function you can use to reference existing resources but not sure that solves where the hooks are deployed

aurelius0523: Which `function name` are you referring to? Even if we allow `HookStack` to be created with different stack name (it is now hardcoded to [aws-codedeploy-hooks](https://github.com/aurelius0523/aws-codedeploy-hooks/blob/8a9d620b7ff7fbe76112f7c96c06f51e36bb2cb3/packages/infra/src/constructs/stack.ts#L20)), it will still fail if we try to deploy multiple `HookStack`s because lambda names are not unique. I feel like I am missing something here

Also, sorry that I am commenting after work hours, please enjoy your evening! 😄 

image

AaronMoat: Yeah, those two functions. 

I'm wondering if we could consider merging this as-is (for this thread anyway), with a comment like 

```ts
// TODO: If deploying multiple stacks in one AWS account, deploy HookStack centrally rather than here
```

(Somewhat per @72636c's [comment](https://github.com/seek-oss/skuba/issues/1640#issuecomment-2323854827))

samchungy: I think just purely because it reduces the complexity of the template I'm happy to merge as is

template/lambda-sqs-worker-cdk/package.json
```diff
@@ -16,6 +16,8 @@
...
+    "@seek/aws-codedeploy-infra": "^2.1.0",
```
samchungy, line 20: Should this be a dev dep?

template/lambda-sqs-worker-cdk/infra/index.ts
aurelius0523: @AaronMoat, @samchungy. I have added that comment and also created a changeset. Using changeset for the first time. Let me know if I am not doing that right

template/lambda-sqs-worker-cdk/package.json
aurelius0523: Yep, moved them!

.changeset/shiny-parents-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': minor
+---
+
+Use aws-codedeploy-hook for lambda-sqs-worker-cdk template
```
AaronMoat, line 5:
```suggestion
template/lambda-sqs-worker-cdk: Replace custom hooks with `@seek/aws-codedeploy-infra` 
```

AaronMoat (approved):
Thank you! @samchungy @72636c might be worth a look from you as well

samchungy (approved):

.changeset/shiny-parents-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': minor
+---
+
+template/lambda-sqs-worker-cdk: Replace custom hooks with `@seek/aws-codedeploy-infra`
```
AaronMoat, line 5:
```suggestion
template/lambda-sqs-worker-cdk: Replace custom hooks with `@seek/aws-codedeploy-infra`
```

template/lambda-sqs-worker-cdk/infra/index.ts
```diff
@@ -1,12 +1,12 @@
 import { App } from 'aws-cdk-lib';
 
+import { HookStack } from '@seek/aws-codedeploy-infra';
```
AaronMoat, line 3:
```suggestion
import { HookStack } from '@seek/aws-codedeploy-infra';
import { App } from 'aws-cdk-lib';

```

template/lambda-sqs-worker-cdk/infra/appStack.ts
```diff
@@ -14,7 +13,6 @@ import {
   aws_sqs,
 } from 'aws-cdk-lib';
 import type { Construct } from 'constructs';
-
 import { config } from './config';
```
AaronMoat, line 16:
```suggestion

import { config } from './config';
```

AaronMoat (approved):
I think this should get formatting happy

AaronMoat:
Gonna force merge and hope for the best 🙈

PR #1645, opened 2024-09-02
Fix: Truncate Buildkite Annotation

### LINKED ISSUE #1331: Build Fails When the Buildkite Annotation is Too Long
The maximum body size of a Buildkite annotation is `1 MiB` as per [Buildkite - Creating an annotation](https://buildkite.com/docs/agent/v3/cli-annotate#creating-an-annotation).

When the body size is greater than `1 MiB`, an error occurs (e.g. [backoffice-access Build 1727 Line 2604](https://buildkite.com/seek/backoffice-access-pr/builds/1727#018c2223-575b-4549-bbc7-5ef1cc512250/402-2640))

```text
Error: Command failed with E2BIG: buildkite-agent annotate --context build-associations.association-diff --style warning ...
```

The [Buildkite.annotate](src/api/buildkite/annotate.ts) function could possibly truncate the annotation similar to [buildkite-plugins/junit-annotate-buildkite-plugin L28-L32](https://github.com/buildkite-plugins/junit-annotate-buildkite-plugin/blob/master/hooks/command#L28-L32) (as suggested by @72636c in [Slack](https://seekchat.slack.com/archives/C02NTMZMRQF/p1701916328999139?thread_ts=1701380175.561809&cid=C02NTMZMRQF)) although that may break the annotation anyway or result in an inaccurate annotation.

GeordieEK:
### Description

When a Buildkite annotation is over `1 MiB` an error occurs.

This PR truncates the buildkite annotation to avoid that error, while still logging the full message to the build log.

### Motivation & Context

Resolves #1331 

### Discussion

`MAX_SIZE` could potentially be passed in as optional attribute of `opts` if desired, would improve testability and possibly add some flexibility if max size ever needs to be different for different situations.


.changeset/quick-icons-taste.md
```diff
@@ -0,0 +1,6 @@
+---
+'skuba': patch
+---
+
+Addresses issue: [#1331](https://github.com/seek-oss/skuba/issues/1331)
+Fix buildkite error when annotation length over 1 MiB by truncating and printing to build log instead
```
AaronMoat, line 6:
```suggestion
api: Truncate Buildkite annotations over 1 MiB to resolve `buildkite-agent` crash
```

src/api/buildkite/annotate.test.ts
```diff
@@ -1,9 +1,12 @@
 import * as execModule from '../../utils/exec';
+import { log } from '../../utils/logging';
 
-import { annotate } from './annotate';
+import { MAX_SIZE, TRUNCATION_WARNING, annotate } from './annotate';
 
 const exec = jest.spyOn(execModule, 'exec');
 const hasCommand = jest.spyOn(execModule, 'hasCommand');
+// Mock impl stops printing to console
```
AaronMoat, line 8:
```suggestion
```

src/api/buildkite/annotate.ts
```diff
@@ -44,6 +49,18 @@ export const annotate = async (
     return;
   }
 
+  // Check if the annotation exceeds the maximum size
+  const buffer = Buffer.from(markdown, 'utf-8');
```
AaronMoat, line 53: I might be missing something obvious! What's the reason to use a buffer here?

AaronMoat (commented):
Thanks for tackling this one! It's not a perfect solution (in the sense that cutting text off right at the end might result in strange output formats), but I don't think any solution is going to be perfect, sadly! 

I've left a few suggestions and questions 😁

src/api/buildkite/annotate.ts
GeordieEK: As opposed to `string.length()` or something else?

My understanding is that the buildkite annotation max is based on a byte size rather than a character length, so this is a more accurate comparison and would account for any characters that may be longer than a byte.

AaronMoat: Interesting, I'm not seeing the distinction drawn in https://buildkite.com/docs/agent/v3/cli-annotate

AaronMoat: Also, I think string length should be ok on that front:

```
$ node
> '🙂‍↔️' .length
5
```

GeordieEK: Oh wow! Did not know .length behaved like that.

AaronMoat (approved):
Looks wonderful! I'll fix that failing test in main

src/api/buildkite/annotate.ts
72636c: `String.length` counts UTF-16 code units though. Can we confirm how Buildkite measures its limit?

PR #1647, opened 2024-09-03
Add pnpm command to update snapshot

aurelius0523:
## Overview
Add a new `pnpm test:template:updateSnapshot` script

## What it does
It leverages existing `update-template.sh` and adds a `-u` option to update snapshot. It will update existing snapshots when running tests in `tmp-${template}` dir and copy the snapshots to `skuba/template` dir

## Why
To make it easier to avoid #1646 because I missed it in #1644

## Further thoughts
I initially thought to make `pnpm test:template` `--updateSnapshot` by default but I am unsure of the existing approach. Assuming that snapshots in template are meant as safeguard for template user and not skuba developers, it might be easier to just have a `husky pre-commit` hook to `--updateSnapshot` by default when modifying templates.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -129,6 +129,16 @@ pnpm test:template greeter
 rm -fr ../tmp-greeter
 ```
 
+Snapshots in `templates` can be updated with the following command but note that manual clean up is necessary afterwards
+
+```shell
+# greeter being name of template
+pnpm test:template:updateSnapshot greeter
+
+# clean up temporary sibling directory
+rm -fr ../tmp-greeter
+```
+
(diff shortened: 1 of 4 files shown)


AaronMoat (approved):
I think I was silly and force merged when I shouldn't have. I think I thought they were failing from the fork, but looking at the build again it seems fairly obvious why it failed. 

But: For development this is FANTASTIC, thank you! 

I believe the snapshots for the CDK template are actually intended for the developer (as well); if a developer takes the template, makes changes, and so on - it's nice to get a diff in the tests / PR for what infrastructure is changing. There are of course other ways we could do this (e.g. staging a changeset)

PR #1651, opened 2024-09-09
Revert "deps: semantic-release ^24.0.0"

AaronMoat:
Discussed internally that we'd prefer to keep the node 18 support over the limited benefits that semantic-release gives us. 

Coupling a node upgrade to the lint upgrades means that node 18 users might not get autofixes; running skuba would fail with a node version error (I think?)

### DIFF
--- a/.changeset/renovate-6eb125a.md
+++ b/.changeset/renovate-6eb125a.md
@@ -1,9 +1,0 @@
----
-'skuba': major
----
-
-deps: semantic-release ^24.0.0
-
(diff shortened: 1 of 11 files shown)


72636c (approved):
I think this wouldn't be terrible given the current Node.js 18-compatible release includes https://seek-oss.github.io/skuba/docs/cli/migrate.html#skuba-migrate-node20 as a migration path at least, but generally agree with the principle

PR #1656, opened 2024-09-10
Remove 'skuba release' command

AaronMoat:
I can't be bothered writing a patch for this even, but open to pushback on that 🌶️


samchungy (approved):
✌️ 👋

PR #1668, opened 2024-09-18
Switch to using the scoped npm token

samchungy:
image

image

Testing switching to using a scoped NPM_TOKEN


AaronMoat (approved):
LGTM, should we try a snapshot?

PR #1680, opened 2024-09-25
Clean up `plugins` handling

AaronMoat:
- https://github.com/seek-oss/skuba/pull/1537#discussion_r1772462725
- https://github.com/seek-oss/eslint-config-seek/pull/149/


72636c (approved):
Now this looks like something I can approve

Issue #1332, opened 2023-12-08
Reference Public ECR Base Images in Docker Images


samchungy:
With regards to recent incidents and pending RFCs, we should switch our references to docker images to their ECR Public equivalents to reduce the likelihood of rate limiting.

In addition, pulling from AWS -> AWS should/could be slightly quicker?

Some questions remain around if we should authenticate to gain higher pulling rates, however, that is a question to pose to the build agency team and whether they could do that for us. 

https://docs.aws.amazon.com/AmazonECR/latest/public/public-registries.html#public-registry-auth
https://docs.aws.amazon.com/AmazonECR/latest/public/public-service-quotas.html

Work would include updating our template dockerfiles to reference ECR images, and stretch goal to create an autofix.

samchungy:
Work to auto login to ECR is complete: https://github.com/SEEK-Jobs/build-agent-images/pull/174

PR #1684, opened 2024-09-26
Switch docker image source to Public ECR

samchungy:
To prevent us hitting rate limits when pulling images from Docker Hub we need to pull our base images from public ECR instead. This PR updates the templates and creates an autofix for the references in our Dockerfiles and Docker Compose files to pull from ECR.

Resolves https://github.com/seek-oss/skuba/issues/1332


template/lambda-sqs-worker-cdk/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1.10
 
-FROM --platform=<%- platformName %> node:20-alpine AS dev-deps
+FROM --platform=<%- platformName %> public.ecr.aws/docker/library/node:20-alpine AS dev-deps
```
AaronMoat, line 3: scope creep but should we be ditching --platform stuff given the docker warnings? @72636c

72636c: I'm comfortable with that 🚫🪟

samchungy: Donezo

src/cli/lint/internalLints/upgrade/patches/8.2.1/patchDockerImages.ts
```diff
@@ -0,0 +1,101 @@
+import { inspect } from 'util';
+
+import fg from 'fast-glob';
+import { readFile, writeFile } from 'fs-extra';
+
+import type { PatchFunction, PatchReturnType } from '../..';
+import { log } from '../../../../../../utils/logging';
+
+const DOCKER_IMAGE_REGEX = /^(FROM\s?.*)(\s)(node|python)(:.*)/gm;
+const DOCKER_IMAGE_PLATFORM_REGEX = /^(FROM\s?.*)(--platform=[^\s]+) /gm;
```
AaronMoat, line 10: Hmm wait I think we only want to remove if it's a hardcoded value like arm64?

AaronMoat: Does this help: https://docs.docker.com/reference/build-checks/from-platform-flag-const-disallowed/

AaronMoat (approved):
Looks good tyvm!

PR #1692, opened 2024-09-30
Skuba 9 preamble

AaronMoat:


.changeset/.PREAMBLE.md
```diff
@@ -0,0 +1,9 @@
+Skuba 9 is a reasonably large release, however we're hoping it won't be too much trouble to upgrade.
```
samchungy, line 1:
```suggestion
skuba 9 is a reasonably large release, however we're hoping it won't be too much trouble to upgrade.
```
Ryan doesn't like capitalism

samchungy (approved):

PR #1705, opened 2024-10-10
Remove advice for global installation

samchungy:
I feel like a global installation of skuba isn't required, as `pnpm` natively caches it for us now so I would like to instead update our advice to using the latest version of skuba to initialise new projects.


AaronMoat (approved):
Agree

PR #1715, opened 2024-10-19
Consider indentation in pnpm Dockerfile autofix

samchungy:
This one irks me because vscode behaves differently around Dockerfiles? On this repo it seems to do spaces of 4 but on others I get 2.

https://github.com/seek-oss/skuba/blob/91af3b6a6d503f7c964e03e9cda48adad4e4c44e/template/lambda-sqs-worker/Dockerfile#L13-L17

This is indented by 4 and when I save it on vscode it doesn't format it.

But on this [repo](https://github.com/SEEK-Jobs/indie-graphql-server/blob/121ae87b1b04bec47195b44affed5ace1bff43b6/Dockerfile.dev-deps#L7-L8) there's a mix of 2 indent and 4 indent and when I save it makes it 2 indent.

So this PR decides to check how the other stuff is indented and copies the indentation.

Prettier does not care about Dockerfiles.


AaronMoat (approved):
:)

samchungy:
image

Seems to work on indie-graph

PR #1720, opened 2024-10-29
bump typescript-eslint

tezf:
fixes https://github.com/typescript-eslint/typescript-eslint/issues/10196

### DIFF
--- a/.changeset/famous-donuts-serve.md
+++ b/.changeset/famous-donuts-serve.md
@@ -2,4 +2,6 @@
 'eslint-config-skuba': minor
 ---
 
-bump typescript-eslint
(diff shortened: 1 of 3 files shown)


AaronMoat (changes requested):
Hi Terry, this shouldn't be necessary - the allowed version range should include any 8.x above the requested version

72636c:
I'm okay with this if we want to force through a version bump for repos that aren't keeping up with lock file maintenance, but I don't think we'd rush to cut a `skuba` release for it.

tezf:
Closing this since we cannot release just the plugin without releasing skuba :(

PR #1714, opened 2024-10-17
Patch Pnpm Dockerfile

samchungy:
Corepack seems to always fetch the latest package manager if you don't have a `package.json` present despite us installing a specific version of `pnpm` beforehand. This autofixes our Dockerfiles to always mount the package.json file when running a pnpm command

https://seekchat.slack.com/archives/C07C02YUF9S/p1729154703475989?thread_ts=1729142626.867769&cid=C07C02YUF9S

PR #1735, opened 2024-11-13
skuba migrate node 22

### LINKED ISSUE #1549: Node 22 support
Issue to track when we're ready to support Node 22 https://nodejs.org/en/blog/announcements/v22-release-announce. Add to the list as you find things:

- [x] Promoted from Current to LTS 
- [x] AWS Lambda support, including via CDK
- [x] Serverless supported or migration path to CDK documented 
- [x] `skuba migrate node22` written 
- [x] https://github.com/seek-oss/skuba/pull/1665 restored
- [x] .node-version
- [x] .nvmrc
- [x] package.json @types/node
- [x] package.json engine (this one might be tricky, if we detect you publish packages, we might leave this alone)
- [x] serverless*.y*ml files
- [x] CDK files. (This night be tricky due to naming but potentially scan through infra/*.ts files. Might do a quick scan through other teams repos.
- [x] docker-compose*.yml
- [x] Dockerfile*
- [x] Tsconfig target (TS 5.7 will support ES2024, Node 22 essentially fully supports ES2024 but once again might bail if we think you're publishing a package) https://node.green/#ES2024
- [x] GH Actions workflows
- [x] Handle shas
- [x] Handle `public.ecr.aws/docker/library/node`

zbrydon:
Resolves #1549


src/cli/migrate/nodeVersion/index.ts
```diff
@@ -14,12 +14,45 @@ type SubPatch = (
...
+type VersionResult = {
+  version: string;
+  err: string | undefined;
+};
+
+type RegistryResponse = {
+  'dist-tags': Record<string, string>;
+};
+
+export const getLatestNodeTypes = async (): Promise<VersionResult> => {
```
samchungy, line 26: Hmmm we might not want the latest latest, because Node releases odd number versions as unstable so there may be Node 23 types coming soon. If possible we want to get the latest version 22 types

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -14,12 +14,45 @@ type SubPatch = (
...
+  const url = 'https://registry.npmjs.org/@types/node';
+  const headers = {
+    accept:
+      'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*',
+  };
+
+  try {
+    const fetchResponse = await fetch(url, { headers });
+    const jsonResponse = (await fetchResponse.json()) as RegistryResponse;
+    const latestNode22 = jsonResponse['dist-tags'].node22;
```
samchungy, line 37: Does a node22 dist tag exist..?

zbrydon: As it turns out... no 😢

zbrydon: Swapped it for your suggestion of
`npm show @types/node@^22 version --json`

src/cli/migrate/nodeVersion/getNode22TypesVersion.ts
```diff
@@ -0,0 +1,4 @@
+import { execSync } from 'child_process';
+
+export const getNode22TypesVersion = () =>
+  execSync("npm show @types/node@^22 version --json | jq '.[-1]'").toString();
```
AaronMoat, line 4: Maybe we could generalise a bit: 

```suggestion
export const getNodeTypesVersionForMajor = (major: number) =>
  execSync(`npm show @types/node@^${major} version --json | jq '.[-1]'`).toString();
```

src/cli/migrate/index.ts
```diff
@@ -3,7 +3,7 @@ import { log } from '../../utils/logging';
 import { nodeVersionMigration } from './nodeVersion';
 
 const migrations: Record<string, () => Promise<void>> = {
-  node20: () => nodeVersionMigration(20),
+  node22: () => nodeVersionMigration(22),
```
AaronMoat, line 6: I feel like it could be fine to keep 20 around just for purposes of people doing a 2-step upgrade? 🤷 maybe not

src/cli/migrate/nodeVersion/packageJsonChecks.test.ts
zbrydon, line 1: I feel we need to test how this behaves in a file structure, would the best way to do this be by utilising `memfs`?

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -6,44 +6,126 @@ import fs from 'fs-extra';
...
+  {
+    id: 'package-json-1',
+    files: '**/package.json',
+    test: /("@types\/node": ")(\^)?[0-9.]+"/gm,
+    replace: '$1$2<%- version %>"',
+  },
+  {
+    id: 'package-json-2',
+    files: '**/package.json',
+    test: /("engines":\s*{[^}]*"node":\s*">=)(\d+)("[^}]*})(?![^}]*"skuba":\s*{[^}]*"type":\s*"package")/gm,
```
zbrydon, line 98: This regex is doing what `checkSkubaType()` does, thoughts on keeping both or just using the function?

72636c (commented):
Apologies, I owe you a proper review but just sending through a few initial comments first. Thanks a bunch for this 🙏

src/cli/lint/internalLints/upgrade/patches/10.0.0/upgradeNode.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const upgradeNode: PatchFunction = async ({
+  mode,
+}): Promise<PatchReturnType> => {
+  if (process.env.SKIP_NODE_UPGRADE) {
+    return {
+      result: 'skip',
+      reason: 'SKIP_NODE_UPGRADE environment variable set',
+    };
+  }
```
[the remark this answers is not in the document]
zbrydon: You have to include the SKIP_NODE_UPGRADE each time to get it to skip otherwise it will attempt to run the patch. 

And as far as env var vs cli option I am easy either way, but if was forced to pick I'd go for the potentially less discoverable option of a cli option so that people are pestered more into keeping their things updated 🤷

src/cli/migrate/nodeVersion/index.test.ts
```diff
@@ -41,22 +53,30 @@ describe('nodeVersionMigration', () => {
           'plugins:\n  - docker#v3.0.0:\n      image: node:18.1.2-slim\n',
         '.buildkite/pipeline2.yml':
           'plugins:\n  - docker#v3.0.0:\n      image: node:18\n',
+        '.buildkite/pipline3.yml':
```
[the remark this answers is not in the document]
zbrydon: Ah no it wasn't 😆 

And yes we're [matching](https://github.com/SEEK-Jobs/indie-toolbox/tree/master/.buildkite) on all files in the directories `**/.buildkite/*` with the idea of being able to handle cases like [this](https://github.com/SEEK-Jobs/indie-toolbox/tree/master/.buildkite).

src/cli/migrate/nodeVersion/getNode22TypesVersion.ts
```diff
@@ -0,0 +1,6 @@
+import { execSync } from 'child_process';
+
+export const getNode22TypesVersion = (major: number) =>
+  execSync(
+    `npm show @types/node@^${major} version --json | jq '.[-1]'`,
+  ).toString();
```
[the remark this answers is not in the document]
zbrydon: This prompted me to take another look at how I was doing this and I think I've reduced some complexity that I introduced while attempting to get tests working, while doing so I removed both the external dependencies.

AaronMoat:
For my benefit, why are we changing ecmascript target and not lib? Should we change skuba's own tsconfig.json (including the bundled one)?

zbrydon:
Ermmm I don't know of any other reason other than it wasn't on the [list](https://github.com/seek-oss/skuba/issues/1549) 😆

AaronMoat:
@samchungy?

AaronMoat:
Per Sam verbally, I think we should upgrade lib as well 😆 though there's complications for npm packages requiring 22, i'm not sure if we want to change those

zbrydon:
Okay so upgrade the lib as well as the target but under the same conditions, ie. not a package and no serverless versions less than v4.

src/cli/lint/internalLints/upgrade/patches/10.0.0/upgradeNode.ts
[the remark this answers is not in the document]
zbrydon: @AaronMoat @samchungy have you got opinions?

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -60,33 +152,111 @@ const runSubPatch = async (version: number, dir: string, patch: SubPatch) => {
...
-        '<%- version %>',
-        version.toString(),
-      );
+      const unPinnedContents = removeNodeShas(contents);
+
+      if (patch.id === 'serverless') {
+        if (!(await validServerlessVersion())) {
+          return;
+        }
+      }
```
72636c, line 161: Reckon we can flip this around a little: instead of having a bunch of ID checks in this function, have each patch specify this logic itself in a self-contained way?

For example, perhaps we could model this particular scenario as:

```typescript
[
  {
    id: 'serverless',
    files: '**/serverless*.y*ml',
    test: [
      /nodejs\d+.x/gm,
      validServerlessVersion,
    ],
    replace: 'nodejs<%- version %>.x',
  }
]
```

Then `runSubPatch` can still operate as a more generic runner like so:

```typescript
if (patch.test) {
  const tests = Array.isArray(patch.test) ? patch.test : [patch.test];
  const results = await Promise.all(tests.map(test => test instanceof RegExp ? test.test(contents) : test()));

  if (!results.every(Boolean)) {
    return;
  }
}
```

We'll need to extend `replace` as well.

src/cli/migrate/nodeVersion/packageJsonChecks.ts
```diff
@@ -0,0 +1,75 @@
...
+        'Serverless version not supported, please upgrade to 4.x to automatically update serverless files',
+      );
+      return false;
+    }
+  } catch (error) {
+    if (isTypeError(error) || isSyntaxError(error)) {
+      return true;
+    }
+    throw error;
+  }
```
72636c, line 45: An alternative here may be to use Zod (you'll still need to consider whether we need to handle errors from `JSON.parse` though):

```typescript
const result = 
  z.object({
    devDependencies: z.object({
      serverless: z.string(),
    }),
  }).transform((data) => data.devDependencies.serverless).safeParse(JSON.parse(packageJson));
  
if (!result.success) {
  return true;
}

const serverlessVersion = result.data;
```

72636c (commented):
🙌

src/cli/lint/internalLints/upgrade/patches/10.0.0/index.ts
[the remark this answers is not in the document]
AaronMoat: No, this PR https://github.com/seek-oss/skuba/pull/1714/files landed a 9.0.1 patch in the file system, and the change landed in 9.1.0

src/cli/migrate/nodeVersion/getNode22TypesVersion.ts
```diff
@@ -0,0 +1,38 @@
...
+    const response = await fetch('https://registry.npmjs.org/@types/node');
+    if (!response.ok)
+      throw new Error(`Failed to fetch: ${response.statusText}`);
+
+    const json = (await response.json()) as {
+      versions: Record<string, unknown>;
+    };
+    const versions = Object.keys(json.versions)
+      .filter((v) => v.startsWith(`${major}.`))
+      .sort((a, b) => b.localeCompare(a, undefined, { numeric: true }));
```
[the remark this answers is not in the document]
zbrydon: 1. Right it's pretty neat, and yeah I believe so. I started using `semver` for the other things so ended up swapping though 😆 
2. Good point I added a check to the filter

src/cli/migrate/nodeVersion/getNode22TypesVersion.ts
```diff
@@ -0,0 +1,38 @@
...
+  defaultVersion: string,
+): Promise<VersionResult> => {
+  try {
+    const response = await fetch('https://registry.npmjs.org/@types/node');
+    if (!response.ok)
+      throw new Error(`Failed to fetch: ${response.statusText}`);
+
+    const json = (await response.json()) as {
+      versions: Record<string, unknown>;
+    };
```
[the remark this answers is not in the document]
zbrydon: Thanks for the find!!
The response is more structured and allows for easily check if the version is deprecated so I swapped over.

src/cli/migrate/nodeVersion/index.ts
zbrydon: I rejigged things [here](https://github.com/seek-oss/skuba/pull/1735/commits/b3b7e116a6d15b573740338d2ba88d3c471f121a) think it cleans things up a fair bit.

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -6,51 +6,143 @@ import fs from 'fs-extra';
...
+    })
+  | Array<
+      ({ files: string; file?: never } | { file: string; files?: never }) & {
+        test?: RegExp;
+        replace: string;
+        id: string;
+      }
+    >;
+
+const SHA_REGEX = /(?<=node.*)(@sha256:[a-f0-9]{64})/gm;
```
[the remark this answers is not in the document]
zbrydon: Yes I think it is, I had to mess around with the docker file regexen since I think a couple of capture groups were fighting each other, but got it working without the duplicate `SHA_REGEX`.

src/cli/migrate/nodeVersion/packageJsonChecks.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import { validServerlessVersion, validSkubaType } from './packageJsonChecks';
+
+jest.spyOn(log, 'warn');
+
+afterEach(() => {
+  jest.resetAllMocks();
+});
+
+describe('validServerlessVersion', () => {
```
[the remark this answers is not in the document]
zbrydon: Aimed to address a few of the `packageJsonChecks` comments with [this](https://github.com/seek-oss/skuba/pull/1735/commits/32b99c4385e25d20aa077b3fb8664269a0bbc628).

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -60,33 +152,111 @@ const runSubPatch = async (version: number, dir: string, patch: SubPatch) => {
...
+      runSubPatch(
+        { nodeVersion, nodeTypesVersion, ECMAScriptVersion },
+        dir,
+        subPatch,
+      ),
+    ),
   );
 };
 
 export const nodeVersionMigration = async (
```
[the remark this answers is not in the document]
zbrydon: mhmmm "or greater" is interesting hadn't even considered it. I will create another snapshot and re-test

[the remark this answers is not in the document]
zbrydon: Added a check before we start the migration that checks that the current node version is less than the target version.  https://github.com/seek-oss/skuba/pull/1735/commits/a20733ec00c4d519cbbb6120bec90e72329cae17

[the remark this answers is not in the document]
zbrydon: Added an [early exit](https://github.com/seek-oss/skuba/blob/590d84699eb132cdbae11a78a6e01ef0960bc034/src/cli/migrate/nodeVersion/index.ts#L199-L201) that checks if the current version is less than or equal to the target version. Which should allow the migration to fix up any botched manual upgrades while not unintentionally downgrading anyone 😅

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -212,54 +172,54 @@ const writePatchedContents = async ({
...
+    ECMAScriptVersion: string;
+    defaultNodeTypesVersion: string;
+  },
   dir = process.cwd(),
 ) => {
   log.ok(`Upgrading to Node.js ${nodeVersion}`);
   try {
-    const { version: nodeTypesVersion, err } = await getNode22TypesVersion(
+    if (!(await isPatchableNodeVersion(nodeVersion))) {
+      throw new Error('Node version is not patchable');
```
72636c, line 208: Since we seem to log the `.js` in a bunch of other spots:

```suggestion
      throw new Error('Node.js version is not patchable');
```

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -212,54 +172,54 @@ const writePatchedContents = async ({
...
+const upgrade = async (versions: Versions, dir: string) => {
   await Promise.all(
-    subPatches.map((subPatch) =>
-      runSubPatch(
-        { nodeVersion, nodeTypesVersion, ECMAScriptVersion },
-        dir,
-        subPatch,
-      ),
-    ),
+    subPatches(versions).map((subPatch) => runSubPatch(dir, subPatch)),
```
72636c, line 189: I think we have cases where 1 file is operated on by more than 1 subpatch. As such, I wonder if we should just run this whole thing sequentially to avoid possible race conditions (i.e. two subpatches both reading the _initial_ state of the file, then the second patch saves over the first patch).

I figure the nested subpatches are meant to address this, but we can probably waste a few extra milliseconds here in exchange for not having to worry about getting the nesting perfectly right.

src/cli/migrate/nodeVersion/getNodeTypesVersion.ts
```diff
@@ -0,0 +1,56 @@
...
+): Promise<VersionResult> => {
+  try {
+    const response = await npmFetch.json('@types/node', {
+      headers: {
+        Accept:
+          'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*',
+      },
+    });
+
+    const parsedVersions = NpmFetchResponse.safeParse(response);
```
72636c, line 33: Out of curiosity, is this noticeably slow in your testing? The response seems massive 😂

src/cli/migrate/nodeVersion/getNodeTypesVersion.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    const version = Object.values(parsedVersions.data.versions)
+      .filter(
+        (v) =>
+          valid(v.version) &&
+          satisfies(v.version, `${major}.x.x`) &&
+          !v.deprecated,
+      )
+      .reduce((a, b) => (gt(a.version, b.version) ? a : b)).version;
```
72636c, line 45: Forgive my nitpick:

```suggestion
    const { version } = Object.values(parsedVersions.data.versions)
      .filter(
        (v) =>
          valid(v.version) &&
          satisfies(v.version, `${major}.x.x`) &&
          !v.deprecated,
      )
      .reduce((a, b) => (gt(a.version, b.version) ? a : b));
```

src/cli/migrate/nodeVersion/getNodeTypesVersion.ts
```diff
@@ -0,0 +1,56 @@
...
+    const response = await npmFetch.json('@types/node', {
+      headers: {
+        Accept:
+          'application/vnd.npm.install-v1+json; q=1.0, application/json; q=0.8, */*',
+      },
+    });
+
+    const parsedVersions = NpmFetchResponse.safeParse(response);
+    if (!parsedVersions.success) {
+      throw new Error('Failed to parse response');
```
72636c, line 35:
```suggestion
      throw new Error('Failed to parse @types/node response from npm');
```

src/cli/migrate/nodeVersion/getNodeTypesVersion.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    return {
+      version,
+    };
+  } catch {
+    return {
+      version: defaultVersion,
+      err: 'Failed to fetch latest @types/node version, using fallback version',
+    };
+  }
```
72636c, line 55: Reckon it's worth logging some info about what failed here, maybe taking something from `catch (err)`?

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+    lte(coercedCurrentVersion, coercedTargetVersion);
+
+  if (!isNodeVersionValid) {
+    log.warn(
+      `Node version in .nvmrc is higher than the target version ${coercedTargetVersion}, please ensure the target version is greater than the current version ${coercedCurrentVersion}`,
+    );
+    return false;
+  }
+
+  log.ok('Valid node version found, proceeding with migration');
```
72636c, line 109:
```suggestion
  log.ok(`Proceeding with migration from Node.js ${coercedCurrentVersion} to ${coercedTargetVersion}`);
```

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  log.ok('skuba type supported, proceeding with migration');
+  return true;
+};
+
+export const isPatchableNodeVersion = async (
+  targetNodeVersion: number,
+): Promise<boolean> => {
+  const currentNodeVersion = await getParentFile('.nvmrc');
```
72636c, line 92: Does anyone have thoughts/data on whether this is a reasonable prerequisite (you must have proper `.nvmrc` that isn't just `lts` in order for the auto-migrate to run)? If not, do we need to fall back to trying to extract a version from a CDK/Docker/Serverless file?

I suspect it may prove difficult for package-only projects, though auto-migrations are less important there anyway (maintainers need to separately and manually consider when they bump their minimum version anyway).

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (!path) {
+    throw new Error(`${file} not found`);
+  }
+  return fs.readFile(path, 'utf-8');
+};
+
+export const extractFromParentPackageJson = async <T extends ZodRawShape>(
+  schema: z.ZodObject<T>,
+): Promise<z.infer<typeof schema> | undefined> => {
+  const packageJson = await getParentFile('package.json');
```
72636c, line 19: Open question here around monorepo support (e.g. Serverless version could be in a _nested_ `package.json`), but I'm okay with us _not_ supporting as long as we bail gracefully

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+      z.object({
+        devDependencies: z.object({
+          serverless: z.string(),
+        }),
+      }),
+    )
+  )?.devDependencies.serverless;
+
+  if (!serverlessVersion) {
+    log.subtle('Serverless version not found, assuming it is not a dependency');
```
72636c, line 46: Would it be useful to log the `package.json` that we found, if any?

```suggestion
    log.subtle(`Serverless version not found in ${packageJsonRelativePath}, assuming it is not a dependency`);
```

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+  const coercedCurrentVersion = coerce(currentNodeVersion)?.version;
+
+  const isNodeVersionValid =
+    coercedTargetVersion &&
+    coercedCurrentVersion &&
+    lte(coercedCurrentVersion, coercedTargetVersion);
+
+  if (!isNodeVersionValid) {
+    log.warn(
+      `Node version in .nvmrc is higher than the target version ${coercedTargetVersion}, please ensure the target version is greater than the current version ${coercedCurrentVersion}`,
```
72636c, line 104:
```suggestion
      `Node.js version ${coercedCurrentVersion ?? 'unknown'} cannot be migrated to ${coercedTargetVersion}`,
```

I wonder if we should `throw` if `coercedTargetVersion` is nullish

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+    log.subtle('Serverless version not found, assuming it is not a dependency');
+    return true;
+  }
+
+  if (!satisfies(serverlessVersion, '4.x.x')) {
+    log.warn(
+      'Serverless version not supported, please upgrade to 4.x to automatically update serverless files',
+    );
+    return false;
+  }
```
72636c, line 55: Should we support Serverless >4, or reckon that's not worth worrying about here? If not, just a small tweak to wording:

```suggestion
  if (!satisfies(serverlessVersion, '4.x.x')) {
    log.warn(
      `Serverless version ${serverlessVersion} cannot be migrated; use Serverless 4.x to automatically migrate Serverless files',
    );
    return false;
  }
```

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (type === 'package') {
+    log.warn(
+      'skuba type package is not supported, packages should be updated manually to ensure major runtime deprecations are intended',
+    );
+    return false;
+  }
+
+  log.ok('skuba type supported, proceeding with migration');
```
72636c, line 85:
```suggestion
  log.ok(`Proceeding with migration of skuba project type ${type}`);
```

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  if (!satisfies(serverlessVersion, '4.x.x')) {
+    log.warn(
+      'Serverless version not supported, please upgrade to 4.x to automatically update serverless files',
+    );
+    return false;
+  }
+
+  log.ok('Serverless version is supported, proceeding with migration');
```
72636c, line 57:
```suggestion
  log.ok(`Proceeding with migration of Serverless version ${serverlessVersion}`);
```

72636c (approved):
🚀

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  if (!type) {
+    log.warn(
+      "skuba type couldn't be found, please specify the type of project in the package.json, to ensure the correct migration is applied",
+    );
+    return false;
+  }
+  if (type === 'package') {
+    log.warn(
+      'skuba type package is not supported, packages should be updated manually to ensure major runtime deprecations are intended',
```
72636c, line 80:
```suggestion
      'Migrations are not supported for packages; update manually to ensure major runtime deprecations are intended',
```

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,111 @@
...
+        skuba: z.object({
+          type: z.string(),
+        }),
+      }),
+    )
+  )?.skuba.type;
+
+  if (!type) {
+    log.warn(
+      "skuba type couldn't be found, please specify the type of project in the package.json, to ensure the correct migration is applied",
```
72636c, line 74:
```suggestion
      "skuba project type not found in ${packageJsonRelativePath}`; add a package.json#/skuba/type to ensure the correct migration can be applied",
```

Though I'm not sure we actually document the `type`s anywhere 🤔, might be something we do later

src/cli/migrate/nodeVersion/checks.ts
zbrydon:
```
└── repo-one
    ├── package.json (A)
    ├── serverless.yaml (A)
    └── apps
        └── foo
            ├── serverless.yaml (B)
            └── package.json (B)
 ```
I think in this example the check for `serverless.yaml (A)` is performed on the `package.json (A)` file and similarly for `serverless.yaml (B)` is checked against `package.json (B)` which I believe is 👌 provided serverless is a dependency in both. 
 
 ```
└── repo-one
    ├── package.json (A)
    ├── serverless.yaml (A)
    ├── tsconfig.json (A)
    └── apps
        └── foo
            ├── tsconfig.json (B)
            └── package.json (B)
 ```
 But in this example if the serverless dependency is in `package.json (B)` and not `package.json (A)` then neither of the `B` files will have engine, target or lib fields upgraded but `A` will.

File: src/cli/migrate/nodeVersion/checks.ts (hunk 23 above)
Comment (zbrydon) [reply src/cli/migrate/nodeVersion/checks.ts]: Yeah admittedly should have called out this fairly large assumption, and I am not sure if it's a valid one.

File: src/cli/migrate/nodeVersion/index.ts (hunk 17 above)
Comment (zbrydon) [reply src/cli/migrate/nodeVersion/index.ts:189]: Yeah nice, that cuts out some unnecessary complexity and a few milliseconds doesn't make a significant difference to a already fairly inefficient process 😆

File: src/cli/migrate/nodeVersion/checks.ts (hunk 23 above)
Comment (72636c) [reply src/cli/migrate/nodeVersion/checks.ts]: Any thoughts @AaronMoat @samchungy 🙂

Comment (AaronMoat) [reply src/cli/migrate/nodeVersion/checks.ts]: I'm personally happy with this pre-req!

Comment (zbrydon) [reply src/cli/migrate/nodeVersion/checks.ts]: I pestered @samchungy at lunch yesterday and it was suggested that we check for a valid version in the `.nvmrc` or `.node-version` files or `engines` field in the `package.json`.

File: src/cli/migrate/nodeVersion/getNodeTypesVersion.ts [hunk 32]
```diff
@@ -28,27 +32,28 @@ export const getNodeTypesVersion = async (
...
     return {
       version,
     };
-  } catch {
+  } catch (err) {
+    log.warn(err);
     return {
       version: defaultVersion,
-      err: 'Failed to fetch latest version, using fallback version',
+      err: 'Failed to fetch latest @types/node version, using fallback version',
```
72636c, line 56: If `defaultVersion` isn't logged elsewhere:

```suggestion
      err: `Failed to fetch latest @types/node version, using fallback version ${defaultVersion}`,
```

src/cli/migrate/nodeVersion/checks.ts
```diff
@@ -0,0 +1,122 @@
...
+export const extractFromParentPackageJson = async <T extends ZodRawShape>(
+  schema: z.ZodObject<T>,
+) => {
+  const { fileContent: packageJson, path } =
+    await getParentFile('package.json');
+  let rawJSON;
+  try {
+    rawJSON = JSON.parse(packageJson) as unknown;
+  } catch {
+    throw new Error('package.json is not valid JSON');
```
72636c, line 31:
```suggestion
    throw new Error(`${path} is not valid JSON`);
```

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -6,131 +6,122 @@ import fs from 'fs-extra';
...
+  {
+    files: '**/package.json',
+    regex: /(\\?"@types\/node\\?": \\?")(\^)?[0-9.]+(\\?(",?)\\?n?)/gm,
+    tests: [isPatchableServerlessVersion],
+    replace: `$1$2${nodeTypesVersion}$4`,
+  },
+  {
+    files: '**/package.json',
+    regex:
+      /(\\?"engines\\?":\s*{\\?n?[^}]*\\?"node\\?":\s*\\?">=)(\d+)\\?("[^}]*})(?![^}]*\\?"skuba\\?":\s*{\\?n?[^}]*\\?"type\\?":\s*\\?"package\\?")/gm,
```
72636c, line 91: Can we improve on`(\d+)` to handle semver ranges in `engines`? Claude suggested `\d+(?:\.\d+)*`

(Technically the field supports more complex expressions like `>=18.18.0 <19.0.0`, but at that point we should eject from regex and just do it in code. Possible in this case as it's easy for us to parse and round-trip, but probably not worth it.)

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -6,131 +6,122 @@ import fs from 'fs-extra';
...
     files: '.node-version*',
-    test: /(v)?\d+\.\d+\.\d+(.+)?/gm,
-    replace: '$1<%- version %>$2',
+    regex: /(v)?\d+\.\d+\.\d+(.+)?/gm,
+    replace: `$1${nodeVersion}$2`,
+  },
+
+  {
+    files: '**/package.json',
+    regex: /(\\?"@types\/node\\?": \\?")(\^)?[0-9.]+(\\?(",?)\\?n?)/gm,
```
72636c, line 84: General question, do we need all these `\\?`s? The patterns seems to work without them but maybe I'm missing something obvious

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -6,131 +6,122 @@ import fs from 'fs-extra';
...
+    regex:
+      /(image: )(public.ecr.aws\/docker\/library\/)?(node:)[0-9.]+(\.[^- \n]+)?(-[^ \n]+)?$/gm,
+    replace: `$1$2$3${nodeVersion}$5`,
   },
   {
-    id: 'node-version',
     files: '.node-version*',
-    test: /(v)?\d+\.\d+\.\d+(.+)?/gm,
-    replace: '$1<%- version %>$2',
+    regex: /(v)?\d+\.\d+\.\d+(.+)?/gm,
```
72636c, line 78: Is support for partial matching like `20.5` required? https://github.com/shadowspawn/node-version-usage?tab=readme-ov-file#compatibility-testing

(`\d+(?:\.\d+)*` may work here too)

src/cli/migrate/nodeVersion/index.ts
```diff
@@ -6,131 +6,122 @@ import fs from 'fs-extra';
...
+    regex: /nodejs\d+.x/gm,
+    tests: [isPatchableServerlessVersion],
+    replace: `nodejs${nodeVersion}.x`,
   },
   {
-    id: 'serverless',
     files: '**/serverless*.y*ml',
-    test: /nodejs\d+.x/gm,
-    replace: 'nodejs<%- version %>.x',
+    regex: /node\d+/gm,
```
72636c, line 54: For these very loose patterns, should we surround with word boundaries `\b` just as a bit of paranoia?

src/cli/migrate/nodeVersion/checks.ts
zbrydon: There was also the suggestion of checking the version of every subPatch that gets applied instead of checking before the subPatches are run. This would require a fair bit of refactoring so opted for just adding those extra checks to `isPatchableNodeVersion`.

72636c (approved):

src/cli/migrate/nodeVersion/index.ts
zbrydon: They exist because I am matching on the stringified version of the `package.json` 😢

src/cli/migrate/nodeVersion/index.ts
zbrydon: I updated the regex to handle the semver ranges like you suggested thank you🙏 
And yeah I agree it would make things significantly nicer but at this point would be a bit of effort to dump the regex 😆

src/cli/migrate/nodeVersion/index.ts
72636c: Hmm, are we double-encoding the `package.json`? If we can avoid that then we should be fine:

```javascript
> /"hello"/.exec('{"hello": "there"}')
[ '"hello"', index: 1, input: '{"hello": "there"}', groups: undefined ]
```

zbrydon: It was my own tests that were fighting me... 😅

PR #1768, opened 2025-01-15
Add names to ESLint config objects

72636c:
https://eslint.org/docs/latest/use/configure/configuration-files#configuration-objects


zbrydon (approved):
Ah, this would have been useful 😅 🙏

PR #1769, opened 2025-01-15
RFC: Try to remove ESLint `languageOptions`

72636c:
We have more modern defaults specified in `eslint-config-seek/base`.

- https://github.com/seek-oss/eslint-config-seek/blob/b06e07d6370c5a0bd585a6dc30ffd7708ff68e06/base.js#L120-L121
- https://github.com/seek-oss/eslint-config-seek/blob/b06e07d6370c5a0bd585a6dc30ffd7708ff68e06/base.js#L138-L141

This may require some manual testing to ensure it doesn't break things on a typical back-end repository that still uses CommonJS.

### DIFF
--- a/.changeset/wet-owls-attack.md
+++ b/.changeset/wet-owls-attack.md
@@ -1,0 +1,10 @@
+---
+'eslint-config-skuba': minor
+---
+
+Revert to modern JavaScript language option defaults
+
+- `ecmaVersion: 5 => latest`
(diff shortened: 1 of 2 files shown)


72636c:
I gave this a whirl on a few repos (one with a package, one with Lambda functions, one service monorepo) and it seems fine 🤷

zbrydon (approved):

PR #1778, opened 2025-02-02
Try to fix build

AaronMoat:


.github/workflows/validate.yml
```diff
@@ -56,7 +56,9 @@ jobs:
           node-version: 20.x
 
       - name: Set up pnpm
-        run: corepack enable pnpm && corepack install
+        run: npm install --global corepack@0.31.0 && corepack enable pnpm && corepack install
+        env:
+          COREPACK_DEFAULT_TO_LATEST: 0
```
72636c, line 61: Does this work _without_ the `npm install`?

If not, I think `COREPACK_DEFAULT_TO_LATEST` may need to be hoisted to the root `env` because it may come into play whenever the `pnpm` shim is invoked 🤔

AaronMoat: No, it didn't (that was my first commit)

72636c: 👍

Ideally `COREPACK_DEFAULT_TO_LATEST` should work by itself to prevent this from happening again in future (else we don't need it at all), then the Corepack upgrade is a separate piece that lets us upgrade to the _current_ new key and latest pnpm

template/oss-npm-package/.github/workflows/release.yml
```diff
@@ -28,7 +31,7 @@ jobs:
           node-version: 20.x
 
       - name: Set up pnpm
-        run: corepack enable pnpm && corepack install
+        run: npm install --global corepack@0.31.0 && corepack enable pnpm && corepack install
```
72636c, line 34: Does Renovate catch these?

.github/workflows/release.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+env:
+  COREPACK_DEFAULT_TO_LATEST: 0
+
```
72636c, line 10: Putting this above `permissions` only in this file is evil work

.changeset/healthy-snakes-attack.md
72636c, line 1: Thanks!

Are you thinking we'll leave the other templates to later?

72636c (approved):

template/oss-npm-package/.github/workflows/release.yml
AaronMoat: Probably not! I'll raise an issue for templates & this to review once things settle

AaronMoat: https://github.com/seek-oss/skuba/issues/1779

.changeset/healthy-snakes-attack.md
AaronMoat: https://github.com/seek-oss/skuba/issues/1779

PR #1789, opened 2025-02-12
Undo corepack hacks, upgrade skuba and templates to node 22

### LINKED ISSUE #1707: Plan for Corepack Removal
https://github.com/nodejs/package-maintenance/pull/606

At the moment we use it to manage our pnpm version.

https://github.com/seek-oss/skuba/blob/138129daab25e612d650b2e79b69fa5386035149/template/lambda-sqs-worker/Dockerfile#L6

https://github.com/seek-oss/skuba/blob/138129daab25e612d650b2e79b69fa5386035149/template/lambda-sqs-worker/package.json#L41

We probably need to write an autofix of sorts but I'm not entirely sure what the best approach would be right now:

`npm i -g corepack && corepack enable`?

That would leave us with installing an unpinned corepack version 🤔.

Just leaving a thread here so we aren't surprised when it pops up in the next major node version

AaronMoat:
Resolves #1707, leaving #1778 open for more permanent changes

Also dropping serverless lol


AaronMoat:
Immediate failure 🚀

.changeset/lucky-kings-shake.md
```diff
@@ -0,0 +1,7 @@
+---
+'skuba': minor
```
AaronMoat, line 2: major? Just to promote it

template/private-npm-package/tsconfig.json
samchungy, line 1: This would force package consumers to use node 22 which is probably not ideal?

AaronMoat: I can roll tsconfig changes back, I was a bit confused why it was in https://github.com/seek-oss/skuba/pull/1735

samchungy: https://node.green/

Is kinda useful. It tells you what more version is compatible with what es version

Generally we would want to support anything in LTS

samchungy: But for things we control like APIs, or workers we can compile to that safely because we're the only consumer

.changeset/lucky-kings-shake.md
AaronMoat: nah

72636c (approved):

PR #1797, opened 2025-02-24
Remove source url tag from gantry values

bridgetbarnes:
https://backstage.myseek.xyz/docs/default/component/gantry/v1/reference/resources/service/#default-tags
https://backstage.myseek.xyz/docs/default/component/gantry/v1/reference/resources/service/#default-tags

Source URL is a default value applied by Gantry, so it doesn't need to be passed in via values


.changeset/happy-seals-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': minor
+---
+
+Remove source URL tag from gantry files
```
AaronMoat, line 5:
```suggestion
template/*-rest-api: Remove `seek:source:url` tag from gantry files
```

AaronMoat (approved):
Thanks!

Issue #1454, opened 2024-02-22
Creating a custom template with  `.png` file breaks BYO templates


relientm96:
## Description

I was trying to create a template with `image/.png` files in the README, but came across this error:

<img width="814" alt="image" src="https://github.com/seek-oss/skuba/assets/37989130/8fc3c29c-a898-40df-a9dc-abbe45d24941">

I assume that there is a `<%` somewhere in the raw version of the `.png` file but it isn't ignored by `skuba` when running `skuba configure`

I also realised that it does not run `pnpm install` neither does it also initialise your project as a git repository. So I think the configure step just fails.

AaronMoat:
Is the raw file safe to attach here in OSS @relientm96?

relientm96:
> Is the raw file safe to attach here in OSS @relientm96?

Yeap the file is safe to be shared but unsure how to share the actual raw contents

Here is a link to the repo that I was trying to `skuba configure` that has the image: https://github.com/SEEK-Jobs/ap-2024-worker-template/blob/master/images/image.png

AaronMoat:
Yep:

```
$ cat Downloads/image.png | grep '<%'
Binary file (standard input) matches
```

relientm96:
> Yep:
> 
> ```
> $ cat Downloads/image.png | grep '<%'
> Binary file (standard input) matches
> ```

Ahh should have though of that, thanks for checking @AaronMoat 😆

AaronMoat:
Any thoughts on resolution? Skip the file formatting if it's in an exclusion list of file formats? Or perhaps inclusion list?

relientm96:
> Any thoughts on resolution? Skip the templating if it's in an exclusion list of file formats? Or perhaps inclusion list?

Some sort of exclusion input on `skuba-configure` would be great as users of skuba would assume everything is included for now.

PR #1808, opened 2025-03-10
Log, instead of crashing, on bad template files

AaronMoat:
Resolves https://github.com/seek-oss/skuba/issues/1454

### DIFF
--- a/.changeset/floppy-cycles-beg.md
+++ b/.changeset/floppy-cycles-beg.md
@@ -1,0 +1,11 @@
+---
+'skuba': patch
+---
+
+init: skip malformed template files
+
+`skuba init` runs templates, either bundled or [BYO](https://seek-oss.github.io/skuba/docs/templates/byo.html), through a series of string templating and processing steps.
+
+Occasionally, binary files can include substrings that appear to be directives for skuba to translate the file contents, which may then proceed to crash.
+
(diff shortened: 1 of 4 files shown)


src/utils/copy.ts
```diff
@@ -52,8 +52,15 @@ interface CopyFilesOptions {
 
 export const createEjsRenderer =
   (templateData: Record<string, unknown>): TextProcessor =>
-  (contents) =>
-    ejs.render(contents, templateData);
+  (sourcePath: string, contents) => {
+    try {
+      return ejs.render(contents, templateData, { strict: false });
+    } catch (err) {
```
AaronMoat, line 58: Debating being smarter with something like https://www.npmjs.com/package/isbinaryfile or just leaving as-is. Thoughts?

72636c: I'm fine with this as is 👍

.changeset/floppy-cycles-beg.md
```diff
@@ -0,0 +1,11 @@
+---
+'skuba': patch
+---
+
+init: skip malformed template files
```
72636c, line 5:
```suggestion
init: Skip malformed template files
```

72636c (approved):
🙇

PR #1830, opened 2025-03-18
Require latest eslint-config-seek 

AaronMoat:
This should make skuba upgrades be seamless, hopefully.


AaronMoat:
https://github.com/un-ts/eslint-plugin-import-x/issues/250

zbrydon (approved):

PR #1832, opened 2025-03-18
Stop throwing on package.json not found

zbrydon:


.changeset/swift-results-feel.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+**migrate:** Remove throw on not found `package.json`
```
AaronMoat, line 5:
```suggestion
migrate: Remove throw on not found `package.json`
```

The bolds get injected by changesets

AaronMoat (approved):

.changeset/swift-results-feel.md
zbrydon: ohhhhh 🪄

tadhglewis (approved):
lgtm. What does this practically do? It just logs something instead of exit 0 when running a command?

zbrydon:
> lgtm. Lazy question, what does this practically do? It just logs something instead of exit 0 when running a command?

It gets used to check if a patch (upgrading a node version) should proceed or not, [eg](https://github.com/seek-oss/skuba/blob/aec6a83f9b058a78c038c077bf17ee49207279f4/src/cli/migrate/nodeVersion/index.ts#L44-L49)

PR #1839, opened 2025-03-26
Clean up logging

AaronMoat:
I missed a log in the bad codeowners check, which meant it failed your lint without any message 😬


samchungy (approved):
danke

72636c (approved):

Issue #1806, opened 2025-03-10
Move pnpm configuration from `.npmrc` to `pnpm-workspace.yaml`


72636c:
npm 12 will disallow third-party options in `.npmrc`. I think this is good for us because we no longer have multiple `.npmrc`s floating around and causing confusion in CI.

https://github.com/orgs/pnpm/discussions/9037

PR #1849, opened 2025-04-07
Migrate pnpm settings from .npmrc to pnpm-workspace

trevorgk:
Resolves https://github.com/seek-oss/skuba/issues/1806


src/cli/lint/internalLints/upgrade/patches/10.1.0/tryClearNpmrcManagedSection.ts
```diff
@@ -0,0 +1,94 @@
...
+  const after = clearManagedSection(before);
+
+  if (before === after) {
+    return {
+      result: 'skip',
+      reason: 'no managed section to clear',
+    };
+  }
+
+  if (mode === 'lint') {
```
AaronMoat, line 75: Worth double checking logging (I think this won't log anything, same for format)

AaronMoat:
This looks really good! I need to stare at the migration docs a bit more. 

I'm wondering if we need to patch more aggressively. The realms of possibilities are:

- The long warnings in .gitignore/.dockerignore
- If they have any extra customisations in .npmrc
- buildkite / dockerfiles that reference .npmrc and perhaps need to instead/as well reference .npm

@72636c @samchungy: any opinions on whether we keep empty .npmrc files around? I lean to yes just for a bit of extra safety on accidentally committing secrets, though unsure if that's genuinely helpful.

trevorgk:
@AaronMoat I erred on the side of keeping them in, in their emptied state. As I felt we could possibly do it in two phases: 
1. Migrate settings from npmrc
2. Potentially tidy up and remove the empty npmrc at a later stage.

72636c:
@AaronMoat without thinking about it a tonne, should we re-add `.npmrc` to `.gitignore` in that typical case where the file would be empty? I think that would do a better job of reducing accidents.

trevorgk:
@72636c off the top of your head, what accidents do you foresee?

72636c:
@trevorgk it doesn't eliminate things like the `private-npm` plugin writing over it and the resulting file & secret being copied and baked in to a container image, or an engineer doing the same by running `cp ~/.npmrc .npmrc` on their local.

docs/deep-dives/pnpm.md
```diff
@@ -65,17 +65,18 @@ pnpm allows us to specify dependencies to hoist via command line or [`.npmrc`].
...
-public-hoist-pattern[]="tsconfig-seek"
+packageManagerStrictVersion: true
+publicHoistPattern:
+  - '@types*'
+  - '*eslint*'
+  - '*prettier*'
+  - esbuild
+  - jest
+  - tsconfig-seek
 # end managed by skuba
```
72636c:
```suggestion
  # end managed by skuba
```

src/cli/lint/internalLints/upgrade/patches/10.1.0/migrateNpmrcToPnpmWorkspace.ts
```diff
@@ -0,0 +1,182 @@
...
+};
+
+const fixDockerfiles = async () => {
+  const fileNames = await glob(['**/Dockerfile*']);
+
+  await Promise.all(
+    fileNames.map(async (fileName) => {
+      const contents = await fs.readFile(fileName, 'utf8');
+      const patched = contents.replaceAll(
+        /--mount=type=bind,source=.npmrc,target=.npmrc/g,
```
72636c, line 67: Do we need the regex here?

```suggestion
        '--mount=type=bind,source=.npmrc,target=.npmrc',
```

72636c (approved):
Apologies for the delay, this looks ace 🚀

> Release notes to tell monorepo setups without skuba-root linting in CI to run format locally and/or manually apply the changes

I am fine with this approach, happy to help wordsmith if required. We can always revisit if we see enough demand to patch those setups automatically.

AaronMoat (approved):
Once Ryan's happy on a re-review of the subsequent commits I reckon this is a merge-and-release situation to avoid skuba 11-specific docs being public without skuba 11 being out.

AaronMoat:
Snapshotting is broken but can't figure out why 😕 https://github.com/seek-oss/skuba/actions/runs/14951052075/job/42000618578

72636c (approved):
👏

.changeset/pink-jobs-thank.md
```diff
@@ -13,3 +13,6 @@ As part of this change:
 - **skuba** will attempt to migrate references to `.npmrc` in Buildkite pipelines and Dockerfiles
 
 **skuba** may not be able to correctly migrate all projects. Check your project for changes that may have been missed, review and test the modified code as appropriate before releasing to production, and [open an issue](https://github.com/seek-oss/skuba/issues/new) if your project files were corrupted by the migration.
+
+**Important**: Monorepos with setups which do not run `skuba lint` in the workspace root will need manual action for this change.
```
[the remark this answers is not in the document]
trevorgk: Missed this, but yes it makes sense, thanks @AaronMoat

PR #1880, opened 2025-05-04
GitHub.putIssueComment: Add comment deletion mode

AaronMoat:


72636c (approved):

samchungy:
I guess my only gripe here is that it's hard to know if the bot is broken which is why we often include the commit id in persistent comments

AaronMoat:
@samchungy It should fail the build right?

samchungy:
Would you want to block the build based on a issue comment though? 🤷‍♂️

AaronMoat:
Depending on the importance, I see the options as 

- Always comment, fail the build. Most robust of "I know it's working"
- Always comment, suppress the build failure. Relies someone noticing the comment is missing or stale. 
- Conditionally comment, suppress the build failure. Relies someone noticing the comment is missing AND that they expect one. 
- Conditionally comment, fail the build. 

I see all as valid TBH depending on the circumstance

PR #1887, opened 2025-05-15
Respect Prettier ignore files in parent directories

72636c:


AaronMoat:
Do we have a gap here where, if we have _multiple_ .prettierignores, context of where you run now matters?

- If you run more nested than the deepest one, you'll only read that.
- If you run at the root, you'll use both.

samchungy (approved):

samchungy:
Going to move to oxfmt in the near future so not going to bother with this

PR #1894, opened 2025-05-23
Fix Node16 module compatibility

samchungy:
Looks like when you do a lock file maintenance on the Octokit deps the type issues you see in other repos disappear as they properly added `type: 'module'` to their package jsons.

This one is the only thing that complains in my project.

```bash
node_modules/.pnpm/skuba@11.0.0_@babel+core@7.27.1_@jest+transform@29.7.0_@swc+core@1.11.29_@typescript-es_d151c735ea38d56f044e010092928178/node_modules/skuba/lib/api/github/pullRequest.d.ts(1,30): error TS1541: Type-only import of an ECMAScript module from a CommonJS module must have a 'resolution-mode' attribute.
```

### DIFF
--- a/.changeset/common-corners-fall.md
+++ b/.changeset/common-corners-fall.md
@@ -2,4 +2,4 @@
 'skuba': patch
 ---
 
-lint,test: Fix `Node16` module resolution compatibility
+lint, test: Fix `Node16` module resolution compatibility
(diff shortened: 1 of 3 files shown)


72636c (approved):
Ooh nice

PR #1897, opened 2025-05-23
Nested repo support for auto commits

zhua633:
Ran into a problem on HPP, the main cause is `getChangedFiles` is treating every file as [added](https://buildkite.com/seek/indie-hirer-posting-preferences-api/builds/3078#0196f13d-588a-4f06-bae0-09ef479cde3b/340-349) because `WORKDIR` is different from the previous commit

Adding support for pnpm workspace monorepo, main changes:

- `getChangedFiles` checks for files changed from the root path 
- `getChangedFiles` now returns absolute path
- `fs.promises.readFile` reads the absolute instead of the relative path

<details>
<summary>See problem breakdown</summary>

see isomorphic-git `StatusMatrix` [example](https://isomorphic-git.org/docs/en/statusMatrix.html#:~:text=The%20result%20is,of%20the%20entry.), to find that files that unmodified files look like the following with `WORKDIR=1`

```
["d.txt", 1, 1, 1], // unmodified
```

because we were running this from within pnpm workspace `apps/api`, it's detecting a change in `WORKDIR=2` for every single file 🫠 
```
row [ 'apps/api/src/storage/postingPreferences/ruleset.ts', 0,2,0 ]
```
</details>

tested [snapshot](https://buildkite.com/seek/indie-hirer-posting-preferences-api/builds/3156#0196fb6b-b433-4c4f-8426-993219712a19) in HPP

### DIFF
--- a/.changeset/soft-clocks-marry.md
+++ b/.changeset/soft-clocks-marry.md
@@ -1,0 +1,5 @@
+---
+'skuba': patch
+---
+
+GitHub.readFileChanges, GitHub.uploadFileChanges: Allowing reading files from directories which are not the git root
--- a/src/api/git/getChangedFiles.ts
+++ b/src/api/git/getChangedFiles.ts
@@ -1,5 +1,5 @@
 import fs from 'fs-extra';
-import git from 'isomorphic-git';
+import git, { findRoot } from 'isomorphic-git';
 
 import {
   ABSENT,
@@ -50,7 +50,8 @@ export const getChangedFiles = async ({
 
   ignore = [],
 }: ChangedFilesParameters): Promise<ChangedFile[]> => {
-  const allFiles = await git.statusMatrix({ fs, dir });
+  const gitRoot = await findRoot({ fs, filepath: dir });
+  const allFiles = await git.statusMatrix({ fs, dir: gitRoot || dir });
   return allFiles
     .filter(
       (row) =>
--- a/src/api/github/push.test.ts
+++ b/src/api/github/push.test.ts
@@ -168,12 +168,18 @@ describe('readFileChanges', () => {
       deletions: [{ path: 'packages/package/delete-path' }],
     };
 
-    expect(fs.promises.readFile).toHaveBeenCalledWith('some-path', {
-      encoding: 'base64',
-    });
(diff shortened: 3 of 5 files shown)


src/api/github/push.test.ts
```diff
@@ -155,9 +155,9 @@ describe('readFileChanges', () => {
     jest.mocked(fs.promises.readFile).mockResolvedValue('base64-contents');
 
     const result = await readFileChanges('/path/to/repo/packages/package', [
-      { path: 'some-path', state: 'added' },
-      { path: 'another-path', state: 'modified' },
-      { path: 'delete-path', state: 'deleted' },
+      { path: 'packages/package/some-path', state: 'added' },
+      { path: 'packages/package/another-path', state: 'modified' },
+      { path: 'packages/package/delete-path', state: 'deleted' },
```
zhua633, line 160: Modified this test because what if there are two files of the same name under `packages/package1/some-path` and `packages/package2/some-path`? 🤔

src/api/git/getChangedFiles.ts
```diff
@@ -50,7 +50,8 @@ export const getChangedFiles = async ({
 
   ignore = [],
 }: ChangedFilesParameters): Promise<ChangedFile[]> => {
-  const allFiles = await git.statusMatrix({ fs, dir });
+  const gitRoot = await findRoot({ fs, filepath: dir });
+  const allFiles = await git.statusMatrix({ fs, dir: gitRoot || dir });
```
72636c, line 54: Stylistic nitpick:

```suggestion
  const allFiles = await git.statusMatrix({ fs, dir: gitRoot ?? dir });
```

.changeset/soft-clocks-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+GitHub.readFileChanges, GitHub.uploadFileChanges: Allowing reading files from directories which are not the git root
```
72636c, line 5:
```suggestion
GitHub.readFileChanges, GitHub.uploadFileChanges: Allowing reading files from directories which are not the Git root
```

72636c (approved):
👏 thank you!

PR #1911, opened 2025-06-17
Experiment with file extensions

samchungy:


src/cli/lint/internalLints/upgrade/index.test.ts
```diff
@@ -55,36 +55,30 @@ describe('upgradeSkuba in format mode', () => {
       description: 'mock',
     };
 
-    jest.mock(`./patches/0.9.0/index.js`, () => ({ patches: [mockUpgrade] }), {
```
samchungy, line 58: Virtual mocks break when you use the module name mapper

PR #1912, opened 2025-06-17
Fix fs-extra imports

samchungy:
Confused at why this was breaking when I was testing ESM but it did 🤷

The fix was using the default import instead of deconstructing

```bash
SyntaxError: Named export 'writeFile' not found. The requested module 'fs-extra' is a CommonJS module, which may not support all module.exports as named exports.
```


72636c (approved):
🆒

PR #1915, opened 2025-06-18
Enforce file extensions

samchungy:
ESM requires file extensions.

We can reduce how much change lands in a ESM Migration PR by enforcing file extensions while they are still on CJS.


packages/eslint-config-skuba/requireExtensions.js
samchungy, line 1: Ripped out of https://github.com/anza-xyz/eslint-plugin-require-extensions

packages/eslint-config-skuba/requireExtensions.js
```diff
@@ -0,0 +1,126 @@
...
+  if (!value || value.endsWith('.js')) {
+    return;
+  }
+
+  if (value.startsWith('.')) {
+    // Relative import, check if it ends with .js
+    return check(context, node, resolve(dirname(context.getFilename()), value));
+  }
+
+  if (value.startsWith('src')) {
```
samchungy, line 47: Handles our special `src` case. I think handling the others is considered under the too hard bucket.

PR #1917, opened 2025-06-19
Document ESM plan

samchungy:
Add some doco I can link to as I'm rolling out the breaking changes.

Also gives everybody an idea about what the heck is going on


samchungy:
Might need a plan for packages which will be ESM. Especially packages part of monorepos.

Need to look at tsdown for that usecase

AaronMoat (approved):

PR #1923, opened 2025-06-20
Apply @seek/logger destination mocks

samchungy:


template/koa-rest-api/src/config.ts
```diff
@@ -32,7 +32,7 @@ const configs: Record<Environment, () => Omit<Config, 'environment'>> = {
   }),
 
   test: () => ({
-    logLevel: Env.string('LOG_LEVEL', { default: 'silent' }),
+    logLevel: Env.string('LOG_LEVEL', { default: 'debug' }),
```
AaronMoat, line 35: Is it worth just 

```suggestion
    logLevel: 'debug',
```

cos you want the tests not to depend on env too much yeah?

AaronMoat (commented):
I'm not sure if the tests are a bit too verbose now. Do we care about all of those fields in context of those assertions?

.changeset/puny-eagles-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': patch
+---
+
+template/lambda-sqs-worker-cdk: Replace logger spies with logger destination mocks
```
AaronMoat, line 5: should we just do /* like https://github.com/seek-oss/skuba/pull/1930

template/koa-rest-api/src/framework/server.test.ts
```diff
@@ -72,13 +69,15 @@ describe('createApp', () => {
...
       .expect('x-api-version', /.+/);
 
-    expect(logger.error).not.toHaveBeenCalled();
-
-    expect(logger.info).toHaveBeenNthCalledWith(
-      1,
-      expect.objectContaining({ status: 404 }),
-      'Client error',
-    );
+    expect(stdoutMock.calls).toMatchObject([
```
AaronMoat, line 72: TIL toMatchObject works with arrays 😮 I thought you had to do .toEqual([expect.objectContaining({...})]);

.changeset/puny-eagles-chew.md
AaronMoat, line 5:
```suggestion
template/lambda-sqs-worker-cdk: Replace logger spies with [`@seek/logger` destination mocks](https://github.com/seek-oss/logger/blob/master/docs/testing.md)
```

AaronMoat (approved):

template/koa-rest-api/src/framework/server.test.ts
72636c: Yeah this is neat! I'm pleasantly surprised it exhaustively matches array elements, so `[1, 2, 3]` does not match `[1, 2]`

Issue #1312, opened 2023-11-09
Add partial batch handling in lambda templates


samchungy:
https://github.com/middyjs/middy/blob/6c9fcc7fa22fc70a60e7db39a5e43c3d09a79c3f/packages/sqs-partial-batch-failure/index.js#L8

Looks fairly simple to replicate. We can go further and wrap it with our async local storage for an optimal flow

PR #1924, opened 2025-06-20
Add partial batch failure handling

samchungy:
https://docs.aws.amazon.com/prescriptive-guidance/latest/lambda-event-filtering-partial-batch-responses-for-sqs/benefits-partial-batch-responses.html

Resolves https://github.com/seek-oss/skuba/issues/1312


template/lambda-sqs-worker-cdk/src/app.ts
```diff
@@ -27,27 +27,28 @@ export const handler = createHandler<SQSEvent>(async (event, ctx) => {
     }
 
     // Run dependency checks otherwise.
-    logger.debug('Smoke test event received');
+    logger.debug(
+      { awsRequestId: ctx.awsRequestId },
```
samchungy, line 31: Not a fan of needing to do this. I have some ideas.

template/lambda-sqs-worker-cdk/package.json
```diff
@@ -18,7 +18,7 @@
     "@aws-sdk/client-lambda": "^3.363.0",
     "@aws-sdk/client-sns": "^3.363.0",
     "@seek/aws-codedeploy-hooks": "^1.2.0",
-    "@seek/logger": "^10.0.0",
+    "@seek/logger": "10.0.0",
```
samchungy, line 21:
```suggestion
    "@seek/logger": "^10.0.0",
```

template/lambda-sqs-worker-cdk/src/framework/handler.ts
```diff
@@ -23,18 +29,46 @@ const withDatadog = <Event, Output = unknown>(
...
-  withDatadog<Event>((event, ctx) =>
-    loggerContext.run({ awsRequestId: ctx.awsRequestId }, async () => {
-      try {
-        const output = await fn(event, ctx);
+  withDatadog<Event>(async (event, ctx) => {
+    try {
+      const output = await fn(event, ctx);
 
-        logger.debug('Function succeeded');
+      logger.debug({ awsRequestId: ctx.awsRequestId }, 'Function completed');
```
samchungy, line 38: Rephrased to `completed` as this could still complete while returning batch failures.

template/lambda-sqs-worker-cdk/src/framework/logging.ts
```diff
@@ -4,11 +4,16 @@ import createLogger, { createDestination } from '@seek/logger';
...
   awsRequestId: string;
 }
 
-export const loggerContext = new AsyncLocalStorage<LoggerContext>();
+interface RecordContext {
+  sqsMessageId: string;
+}
+
+export const lambdaContext = new AsyncLocalStorage<LambdaContext>();
+export const recordContext = new AsyncLocalStorage<RecordContext>();
```
samchungy, line 16: I decided to add a second context here as I think it's important to correlate logs related to a single record since we are now doing partial batch failure

template/lambda-sqs-worker-cdk/infra/config.ts
```diff
@@ -37,6 +39,7 @@ const configs: Record<Environment, Config> = {
   prod: {
     appName: '<%- serviceName %>',
     workerLambda: {
+      batchSize: 5,
```
72636c, line 42: Very unimportant, but is it worth sticking with the AWS default of 10?

template/lambda-sqs-worker-cdk/src/framework/handler.ts
```diff
@@ -38,3 +44,30 @@ export const createHandler = <Event, Output = unknown>(
...
+  (
+    fn: (record: SQSRecord, ctx: LambdaContext) => Promise<unknown>,
+  ): Handler<SQSEvent, SQSBatchResponse> =>
+  async (event, ctx) => {
+    const processRecord = (
+      record: SQSRecord,
+    ): Promise<SQSBatchItemFailure | undefined> =>
+      recordContext.run({ sqsMessageId: record.messageId }, async () => {
+        try {
+          await fn(record, ctx);
```
72636c, line 58: I won't block this PR as it's an improvement over the current state, but this model doesn't promote optimal I/O usage as you're still processing the records individually within the process, unless you try to have something clever like dataloader behind the scenes. Not sure if that's worth mentioning in the template, or if we'd prefer simplicity to be the default here 🤷

72636c (approved):

PR #1940, opened 2025-07-07
Migrate fully to Zod v4

samchungy:
I updated all the templates but forgot our actual repo 🤡 

Don't think this warrants a changeset.

### DIFF
--- a/src/utils/template.ts
+++ b/src/utils/template.ts
@@ -2,11 +2,6 @@ import path from 'path';
 
 import fs from 'fs-extra';
 import * as z from 'zod/v4';
-import type {
-  $InferInnerFunctionType,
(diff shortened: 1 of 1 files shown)


template/koa-rest-api/src/framework/validation.ts
```diff
@@ -59,15 +60,15 @@ const parseTuples = (
...
   ctx,
   input,
   schema,
 }: {
   ctx: Context;
   input: unknown;
-  schema: z.ZodSchema<Output, Input>;
-}): Output => {
+  schema: T;
+}): z.infer<T> => {
```
samchungy, line 71: The previous approach actually doesn't work for things like Branded types.

72636c (approved):

PR #1943, opened 2025-07-11
Try to fix autofixes with LFS

### LINKED ISSUE #1732: Autofixes modify unchanged LFS files
`isomorphic-git` doesn't have LFS support, and so our `getChangedFiles` logic is treating them as modified (it seems).

AaronMoat:
Resolves #1732

### DIFF
--- a/src/api/git/remote.int.test.ts
+++ b/src/api/git/remote.int.test.ts
@@ -3,7 +3,7 @@ import memfs, { vol } from 'memfs';
 
 import { getOwnerAndRepo } from './remote';
 
-jest.mock('fs-extra', () => memfs);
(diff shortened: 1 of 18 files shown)


src/api/git/remote.int.test.ts
```diff
@@ -3,7 +3,7 @@ import memfs, { vol } from 'memfs';
 
 import { getOwnerAndRepo } from './remote';
 
-jest.mock('fs-extra', () => memfs);
+jest.mock('fs', () => memfs);
```
AaronMoat, line 6: It's not correct to mock fs-extra with memfs, because memfs mimics fs, not fs-extra, and fs-extra uses fs under the hood.

PR #1948, opened 2025-07-15
Introduce custom conditions

samchungy:
I've been playing around with custom conditions which are required to get monorepos with live types to work properly. There's a hack in `indie-metrics-api` I would love to delete.

In doing so, it's definitely the recommended way to get our subpath imports working with our tooling.

1. TypeScript natively supports it ✅ 
2. Vitest natively supports it ✅ 
3. Tsx natively supports it ✅ 

We could potentially pipe it automatically through our `skuba node` wrapper unless this lands soon: https://github.com/privatenumber/tsx/issues/574.

Alternatively we could pipe it through `pnpm`'s `nodeOptions`


docs/deep-dives/esm.md
```diff
@@ -88,20 +88,19 @@ This shouldn't cause any issues, but out of caution, we will release this as a n
...
+We will replace `skuba-dive/register` with subpath imports and custom conditions, a native solution supported by both [TypeScript] and [Node.js].
 
 The subpath imports feature allows us to define custom paths in `package.json`, enabling us to import modules using simplified paths without needing to use deep relative paths.
 
 package.json:
 
 ```diff
 {
-  "name": "my-package",
+  "name": "my-repo",
```
72636c, line 99: I recall someone had an opinion around scoping everything in case of accidental publishing, though we can re-litigate with Security if we hate it

```suggestion
  "name": "@seek/my-repo",
```

docs/deep-dives/esm.md
```diff
@@ -145,9 +145,13 @@ Your local `tsconfig.json` files will require a `baseUrl` and `rootDir` to help
 This allows us to import modules like this:
 
 ```ts
-import { module } from '#src/imported-module';
+import { module } from '#src/imported-module.js';
 ```
 
+[Custom conditions] enable tooling to use TypeScript source files during development while automatically switching to the correct compiled outputs for deployment or publishing. This is done without additional configuration to modify package.json imports and exports. Monorepo users may already be familiar with this concept through pnpm's `publishConfig` feature. For more details, see [Live types in a TypeScript monorepo].
```
72636c, line 151: Lazy question: does tsc rewrite `customConditions` during transpile, or are we now relying on [`node --conditions`](https://nodejs.org/api/cli.html#-c-condition---conditionscondition)?

72636c (approved):

docs/deep-dives/esm.md
72636c: ah right!

PR #1952, opened 2025-07-17
Apparently you gotta double $$

samchungy:
Don't ask me why.

Tested with `pnpm changeset version`

image


zbrydon (approved):
Why?

72636c (approved):
:shipit:

(it's escaping)

PR #1953, opened 2025-07-18
Prevent Jest config and setup files from being linted

samchungy:
Unfortunately Jest's TypeScript processor doesn't use modern standards and can't read TypeScript files with `.js` extensions 🙄. So any `.js` extension it encounters will cause an error.

This change won't stop cases where someone may import another TypeScript file which is using extensions but should prevent a decent majority of cases.

See: https://github.com/SEEK-Jobs/indie-graphql-server/pull/3325/commits/40fdf8a5f150a98e9fe6edd1adf7b2692e18f6f7 as an example.

### DIFF
--- a/packages/eslint-config-skuba/requireExtensions.js
+++ b/packages/eslint-config-skuba/requireExtensions.js
@@ -195,7 +195,9 @@ const requireExtensionsPlugin = {
               fix(fixer) {
                 return fixer.replaceText(
                   node.source,
-                  `'${node.source.value}/index.js'`,
+                  node.source.value.endsWith('/')
+                    ? `'${node.source.value}index.js'`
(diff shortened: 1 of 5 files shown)


packages/eslint-config-skuba/requireExtensions.js
```diff
@@ -195,7 +195,9 @@ const requireExtensionsPlugin = {
               fix(fixer) {
                 return fixer.replaceText(
                   node.source,
-                  `'${node.source.value}/index.js'`,
+                  node.source.value.endsWith('/')
```
samchungy, line 198: My ocd: https://seekchat.slack.com/archives/CDCPCEPV3/p1752793805171339?thread_ts=1752732605.041009&cid=CDCPCEPV3

PR #1969, opened 2025-07-29
RFC: Propose `no-sync-in-promise-iterable`

72636c:
Follows on from seek-oss/eslint-config-seek#227.

This was initially drafted by Claude but I basically rewrote it.

---

Looking for input on:

- [x] Whether the general approach seems reasonable
- [x] Where we should put this (standalone plugin, skuba config, else?)

  Trying my luck with a plugin!

- [ ] Later, testing whether it is prohibitively slow

Documentation: https://github.com/seek-oss/skuba/blob/no-sync-in-promise-iterable/docs/eslint-plugin/no-sync-in-promise-iterable.md

---

~~Note: I haven't actually wired this up to the config yet as will need to sort out transpile, its use of `node:test`, different `module` options, etc 😅~~

<details>

<summary><s>eslint.config.js</s></summary>

```js
module.exports = [
  // ...
  {
    plugins: {
      skuba: {
        rules: {
          'no-sync-in-promise-iterable': require('./no-sync-in-promise-iterable.js'),
        },
      },
    },
    files: ['**/*.ts', '**/*.tsx'],
    languageOptions: {
      parser: require('@typescript-eslint/parser'),
      parserOptions: {
        project: './tsconfig.json',
        tsconfigRootDir: __dirname,
      },
    },
    rules: {
      'skuba/no-sync-in-promise-iterable': 'error',
    },
  },
];
```

</details>


packages/eslint-config-skuba/no-sync-in-promise-iterable.test.ts
```diff
@@ -0,0 +1,365 @@
...
+      code: `Promise.${method}()`,
+    },
+    {
+      code: `Promise.${method}([])`,
+    },
+    {
+      code: `Promise.${method}([,,])`,
+    },
+    {
+      code: `Promise.${method}([1, 2, 3])`,
```
samchungy, line 37: Would it be simpler to just ban any non promise results from being added to these? I don't see any value in:

```ts
Promise.all([1, 2, 3]);
```

or

Even if it's allowed in JavaScript.

PR #1958, opened 2025-07-21
Subpath imports

zhua633:
part of [esm migration](https://github.com/seek-oss/skuba/blob/main/docs/deep-dives/esm.md#2-replace-skuba-diveregister-with-subpath-imports), switching to use `#src/` aliasing instead of `skuba-dive/register`

mostly generated, left comments on manual changes parts

tested snapshots for [monorepo](https://github.com/SEEK-Jobs/indie-hirer-posting-preferences-api/pull/865) and [microservice](https://github.com/SEEK-Jobs/ie-job-categories-api/pull/1744)

PR #1970, opened 2025-07-29
Patch to automate custom conditions

zhua633:
automate some more things: https://github.com/seek-oss/skuba/pull/1958#discussion_r2227356185

tested in monorepo and microservice


src/cli/lint/internalLints/upgrade/patches/12.0.1/configureTsConfigForESM.ts
```diff
@@ -0,0 +1,287 @@
...
+      typeof tsconfigJson.compilerOptions !== 'object'
+    ) {
+      tsconfigJson.compilerOptions = {};
+    }
+
+    const compilerOptions = tsconfigJson.compilerOptions;
+
+    if (compilerOptions.paths !== undefined && !isMonoRepo) {
+      delete compilerOptions.paths;
+    }
```
zhua633, line 184: intended change for microservices:
```diff
-    "paths": {
-      "src": ["src"]
-    },
```

monorepos would keep it:
```
"paths": {
     "src/*": ["apps/api/src/*", "apps/worker/src/*"]
},
```

src/cli/lint/internalLints/upgrade/patches/12.0.1/configureTsConfigForESM.ts
```diff
@@ -0,0 +1,287 @@
...
+
+    const tsconfigJson = parseResult.data;
+
+    if (
+      typeof tsconfigJson.extends === 'string' &&
+      !tsconfigJson.extends.startsWith('skuba/')
+    ) {
+      log.subtle(
+        'Skipping tsconfig.json that does not extend skuba/config/tsconfig.json',
+      );
```
zhua633, line 169: `tsconfig.json` that extends from the project root would not need to be reconfigured

src/cli/lint/internalLints/upgrade/patches/12.0.1/configureTsConfigForESM.ts
```diff
@@ -0,0 +1,287 @@
...
+    return contents;
+  }
+};
+
+export const tryConfigureTsConfigForESM: PatchFunction = async ({
+  mode,
+}): Promise<PatchReturnType> => {
+  const packageJsonPatterns = ['**/package.*json'];
+  const tsconfigJsonPatterns = ['**/tsconfig.*json'];
+  const jestConfigPatterns = ['**/jest.config.*ts'];
```
zhua633, line 211: this would cover for things like `package.base.json`, `jest.config.int.ts`, with the exception that later `globOptions` would ignore `tsconfig.build.json`

.changeset/warm-areas-hunt.md
```diff
@@ -0,0 +1,60 @@
...
+---
+
+**format, lint:** Auto-configures `package.json`, `tsconfig.json` and `jest.config.ts` for ESM migration support.
+
+## Manual changes that may be required
+
+If you get this Jest error:
+
+```
+TypeError: A dynamic import callback was invoked without --experimental-vm-modules
```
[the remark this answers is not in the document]
zhua633: did [it](https://github.com/seek-oss/skuba/pull/1958/commits/33f11858d3fed154e1468383689a673920ba4413) on the other branch oops

samchungy:
This is pretty good so far but I think we can be a little more selective with what `package.json` files and `tsconfig` files we're adding the mapping to.

Can we:

1. Find a root `tsconfig.json` and only add customConditions to that one? We should be able to find this by finding the gitRoot and looking for an adjacent `tsconfig.json`.

2. Find all `tsconfig.json` files with `src` in them.

	If they contain `src`, for every path in them, find the nearest package.json and add the `imports` section to it.
	
	eg.
	
	```ts
	"paths": {
      "src/*": ["apps/api/src/*", "apps/worker/src/*"]
    },
    ```
    
    If `apps/api/` contains a package.json and tsconfig.json add the imports section + rootDir '.'. If there is also a tsconfig.build.json, add a rootDir: 'src' to it.

src/cli/lint/internalLints/upgrade/patches/12.2.0/configureTsConfigForESM.ts
```diff
@@ -0,0 +1,272 @@
...
+    '^#src/(.*)\\.js$': formattedNamesWithPath,
+    '^#src\/(.*)$': formattedNamesWithPath,
+  };
+
+  if (isTypeScriptJestConfig(contents)) {
+    return addModuleNameMapperToTypeScript(contents, moduleNameMapper);
+  }
+
+  try {
+    const parseResult = packageJsonSchema.safeParse(JSON.parse(contents));
```
samchungy, line 91: does this make sense?

zhua633: thanks for fixing this one

PR #1985, opened 2025-08-02
Add ESM migration notes on rewriteRelativeImportExtensions

samchungy:


docs/deep-dives/esm.md
```diff
@@ -160,12 +160,64 @@ We will apply a community codemod to help with the transition, but it will likel
...
++ import { module } from './imported-module.ts';
+```
+
+Imports beginning with `'#src/'` will need to be rewritten without extensions, as the `rewriteRelativeImportExtensions` feature only works for relative imports. This means we would need to update our `package.json` imports:
+
+```diff
+  "imports": {
+    "#src/*": {
+-    "@seek/my-repo/source": "./src/*",
++    "@seek/my-repo/source/*": "./src/*.js",
```
samchungy, line 186:
```suggestion
+    "@seek/my-repo/source/*": "./src/*.ts",
```

docs/deep-dives/esm.md
```diff
@@ -160,12 +160,64 @@ We will apply a community codemod to help with the transition, but it will likel
 
 ---
 
+## Future considerations
+
+TypeScript 5.7 introduced [`rewriteRelativeImportExtensions`], which allows us to rewrite relative import paths to include the `.js` file extension. When combined with [`allowImportingTsExtensions`], this would allow us to import `.ts` files directly.
+
+Node.js also recently added [native type stripping support], which means we could start using Node.js to run TypeScript code directly without compilation.
```
72636c, line 167: FWIW I'm not sure this should be a goal for production applications, especially if/when we go harder on bundling and tree shaking. But it would simplify local run, REPL, scripting, etc.

72636c (approved):
👏

PR #1997, opened 2025-08-08
Auto map experimental vm modules to `pnpm test`

samchungy:
With the move to `"module": "node18",` Jest no longer transforms `await import` calls to require which leaves us in a state where need to enable experimental vm modules or switch them to requires.

I think this is probably easier given we're going to throw Jest in the bin anyway


72636c (approved):
🙈

PR #2002, opened 2025-08-10
Fix `test:template:updateSnapshot` script

72636c:
Discovered this in the process of working on #2001.


scripts/test-template.sh
```diff
@@ -108,7 +108,7 @@ pnpm format
 
 if [ "$update_snapshot" = true ]; then
   echo "--- pnpm test --updateSnapshot ${template}"
-  pnpm test -- --updateSnapshot
```
72636c, line 111: @samchungy too lazy to bisect at the moment, any chance #1997 broke this?

PR #2006, opened 2025-08-11
Add DataDog layer custom regex

samchungy:
Forgot this doesn't extend Rynovate


AaronMoat (approved):
should we be using rynovate for this

PR #2018, opened 2025-08-11
template/lambda-sqs-worker-cdk: Use Datadog runtime layer

72636c:


.changeset/fruity-walls-hear.md
72636c, line 1: Add @samchungy's manifesto here?

template/lambda-sqs-worker-cdk/infra/__snapshots__/appStack.test.ts.snap
```diff
@@ -941,14 +953,26 @@ exports[`returns expected CloudFormation stack for prod 1`] = `
...
         "Layers": [
+          {
+            "Fn::Join": [
+              "",
+              [
+                "arn:aws:lambda:",
+                {
+                  "Ref": "AWS::Region",
+                },
+                ":464622532012:layer:Datadog-Node22-x:126",
```
samchungy, line 972: This might become annoying with Renovates to update

samchungy: Yep, I think other parts of the snap would need to be updated when the runtime gets updated anyway

PR #2024, opened 2025-08-13
Exempt more `skuba/no-sync-in-promise-iterable` scenarios

72636c:


AaronMoat (approved):
I can't say we have the same definition of obvious

PR #2040, opened 2025-08-26
Prioritise user-defined `moduleNameMapper` entries over skuba defaults

samchungy:


samchungy:
Shit this might be impossible to test

PR #2065, opened 2025-09-18
Add `minimumReleaseAge`

zbrydon:
https://github.com/pnpm/pnpm/releases/tag/v10.17.0

### DIFF
--- a/template/base/_pnpm-workspace.yaml
+++ b/template/base/_pnpm-workspace.yaml
@@ -3,10 +3,11 @@ ignorePatchFailures: false
 minimumReleaseAge: 4320 # 3 days
 minimumReleaseAgeExclude:
   - '@seek/*'
-  - 'eslint-config-skuba'
-  - 'eslint-plugin-skuba'
-  - 'skuba'
-  - 'skuba-dive'
+  - eslint-config-seek
+  - eslint-config-skuba
+  - eslint-plugin-skuba
+  - skuba
+  - tsconfig-seek
 packageManagerStrictVersion: true
 publicHoistPattern:
   - '@types*'
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -3,10 +3,11 @@ ignorePatchFailures: false
 minimumReleaseAge: 4320 # 3 days
 minimumReleaseAgeExclude:
   - '@seek/*'
-  - 'eslint-config-skuba'
-  - 'eslint-plugin-skuba'
-  - 'skuba'
-  - 'skuba-dive'
+  - eslint-config-seek
+  - eslint-config-skuba
+  - eslint-plugin-skuba
+  - skuba
+  - tsconfig-seek
 packageManagerStrictVersion: true
 publicHoistPattern:
   - '@types*'


src/cli/lint/internalLints/upgrade/patches/12.2.0/patchPnpmWorkspace.ts
```diff
@@ -0,0 +1,87 @@
...
+    pnpmWorkspaceContents.includes('minimumReleaseAgeExclude')
+  ) {
+    return {
+      result: 'skip',
+      reason: '`minimumReleaseAge` already exists in pnpm-workspace.yaml',
+    };
+  }
+
+  const newContents = `${pnpmWorkspaceContents}
+minimumReleaseAge: 1440
```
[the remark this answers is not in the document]
zbrydon: Yeah open to suggestions, picked a day because that's what was in the pnpm changelog 😆 

Mhmmm didn't think about that, kinda like aligning them 🤔

template/base/_pnpm-workspace.yaml
```diff
@@ -1,4 +1,9 @@
 # managed by skuba
+minimumReleaseAge: 4320
```
72636c, line 2: Not really fussed, but thoughts on:

```suggestion
minimumReleaseAge: 4320 # 3 days
```

template/base/_pnpm-workspace.yaml
```diff
@@ -1,4 +1,9 @@
...
+minimumReleaseAgeExclude:
+  - '@seek/*'
+  - '*skuba*'
+  - '*seek*'
```
72636c, line 6: Is it concerning to exempt package patterns that we don't control (i.e. other than `@seek/*`)?

I'm thinking: not great, but probably not a dealbreaker. `skuba` isn't going to appear in package names organically, so it would require a targeted attack and two steps to exploit (compromise a package `left-pad` to add a benign dependency on a `bad-skuba@^1.0.0` package then wait for 3 days for `left-pad` to be bumped, publish `evil-skuba@1.0.1` with badness included then wait for the next lock file maintenance). `seek` may be more likely but again would likely require a targeted attack.

Not sure I like the alternative of tediously listing out the exclusions, unless we keep it to a pretty small subset (e.g. it's probably fine if `seek-koala` is not updated immediately).

src/cli/configure/processing/configFile.ts
```diff
@@ -1,3 +1,5 @@
+import z from 'zod/v4';
```
72636c, line 1: Should this be

```suggestion
import * z from 'zod';
```

src/cli/configure/processing/configFile.ts
```diff
@@ -41,16 +43,62 @@ export const generateIgnoreFileSimpleVariants = (patterns: string[]) => {
   return set;
 };
 
+const ammendPnpmWorkspaceTemplate = (
```
72636c, line 46:
```suggestion
const amendPnpmWorkspaceTemplate = (
```

src/cli/configure/processing/configFile.ts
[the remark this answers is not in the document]
72636c: Do we have an immediate need for overloads? If not, I wonder if we leave this out (or perhaps undocumented so as to not encourage uptake) until we implement `skuba.config.ts` properly.

.changeset/chubby-tables-hug.md
```diff
@@ -0,0 +1,13 @@
+---
+'skuba': minor
+---
+
+lint: Adds `minimumReleaseAge` and `minimumReleaseAgeExclude` to pnpm-workspace.yaml
```
72636c, line 5:
```suggestion
lint: Add `minimumReleaseAge` and `minimumReleaseAgeExclude` to `pnpm-workspace.yaml`
```

72636c (approved):

template/base/_pnpm-workspace.yaml
zbrydon: Yeah thinking through it more, the motivation was to keep the exclusion list short while not getting in the way of development. But if we keep the ability to overload then that provides an escape hatch if required, and makes the default less greedy. 🤔

template/base/_pnpm-workspace.yaml
zbrydon: Like it saves people having to ask what it means 😆

src/cli/configure/processing/configFile.ts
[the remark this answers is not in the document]
zbrydon: Think yes, only if we end up not excluding packages that we want to publish and test quickly 🤔

I wound it back to just `@seek/*` but am now reconsidering after looking at all the exclusions listed [here](https://github.com/seek-oss/rynovate/blob/master/default.json). 

Feel like it might be good to have them the same so we don't end up with renovates hanging around unable to install...

PR #2070, opened 2025-09-22
RFC: pnpm config dependency

zbrydon:


packages/pnpm-plugin/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": "0.0.1",
+  "description": "PNPM config dependency for skuba that configures recommended PNPM settings",
+  "homepage": "https://github.com/seek-oss/skuba#readme",
+  "bugs": {
+    "url": "https://github.com/seek-oss/skuba/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/seek-oss/skuba.git",
+    "directory": "packages/pnpm-plugin"
```
72636c, line 12: Oh TIL `directory`, we should probably set it for our other packages

packages/pnpm-plugin/pnpmfile.cjs
```diff
@@ -0,0 +1,34 @@
+// @ts-check
+module.exports = {
+  hooks: {
+    /** @param {import("@pnpm/types").PnpmSettings} config */
+    updateConfig(config) {
+      // @ts-expect-error Property isn't in PnpmSettings for some reason
+      config.publicHoistPattern ??= [];
+      // @ts-expect-error Property isn't in PnpmSettings for some reason
+      config.publicHoistPattern.push(
```
72636c, line 9: Should our patterns be _prepended_ so consumers can define overrides with `!`?

https://pnpm.io/settings#publichoistpattern

packages/pnpm-plugin/pnpmfile.cjs
```diff
@@ -0,0 +1,34 @@
...
+        'prettier',
+        'esbuild',
+        'jest',
+        'tsconfig-seek',
+        'typescript',
+      );
+
+      config.ignorePatchFailures = false;
+      // @ts-expect-error Property isn't in PnpmSettings for some reason
+      config.packageManagerStrictVersion = true;
```
72636c, line 22: Similar question on whether stuff like this should be `??=` to allow consumer to override 🤔

72636c (approved):
Seems reasonable enough, though agree with Sam that magic is not ideal 🥲

packages/pnpm-plugin/pnpmfile.cjs
```diff
@@ -0,0 +1,33 @@
...
+      config.publicHoistPattern ??= [
+        '@types*',
+        'eslint',
+        'eslint-config-skuba',
+        'prettier',
+        'esbuild',
+        'jest',
+        'tsconfig-seek',
+        'typescript',
+      ];
```
72636c, line 22: To support merging, the array configs could be like

```suggestion
      (config.publicHoistPattern ??= []).unshift(
        '@types*',
        'eslint',
        'eslint-config-skuba',
        'prettier',
        'esbuild',
        'jest',
        'tsconfig-seek',
        'typescript',
      );
```

Or, if the above is too spooky to look at:

```suggestion
      config.publicHoistPattern = [
        '@types*',
        'eslint',
        'eslint-config-skuba',
        'prettier',
        'esbuild',
        'jest',
        'tsconfig-seek',
        'typescript',
        ...(config.publicHoistPattern ?? []),
      ];
```

72636c: [This](https://github.com/SEEK-Jobs/indie-awsk-ui/pull/349) seems to imply that plugins don't work properly with `pnpm fetch` and `pnpm install --offline` 🤔

PR #2075, opened 2025-09-25
template: Remove syntax directive from `Dockerfile`s

72636c:
Our templated Dockerfiles require frontend version 1.10+ to mount build secrets as environment variables:

- https://docs.docker.com/build/buildkit/dockerfile-release-notes/#1100
- https://redirect.github.com/moby/buildkit/pull/5215

Our local & CI environments should have more recent Dockerfile frontend versions bundled in their Docker toolchains; for example, I'm currently running Docker Engine 28.4.0 locally which should support 1.18.0.

We can remove this directive until we have a need for newer syntax that is not yet built in. This hopefully reduces an online dependency on Docker services at build time.

---

- [ ] Add changeset
- [ ] Determine if we want to patch?! (with a disclaimer that it may break if you use other syntax, and CI should trivially pick that up)


zbrydon (approved):
Are we going to bother removing it for people?

PR #2108, opened 2025-10-15
Handle more Serverless cases

samchungy:
A worker we use is using `includes`.  Thanks Claude.


packages/eslint-plugin-skuba/tsconfig.json
```diff
@@ -7,8 +7,7 @@
     "outDir": "lib",
     "removeComments": false,
     "skipLibCheck": true,
-    "target": "ES2021",
-    "customConditions": ["@seek/skuba/source"]
```
72636c: 🙇

PR #2120, opened 2025-10-21
Use templated values for gantry environment.

trevorgk:
Use templated values to guard against these types of errors during deployments 🙈 
`process.env.DEPLOYMENT is not a supported choice: 'dev'. Expected one of: ['local', 'test', 'mn-dev-ats-relations', 'mn-prod-ats-relations']`

### DIFF
--- a/template/express-rest-api/.gantry/prod.yml
+++ b/template/express-rest-api/.gantry/prod.yml
@@ -1,8 +1,8 @@
 deployment: <%- prodGantryEnvironmentName %>
 env: production
 
-maxInstanceCount: 10
 minInstanceCount: 3
+maxInstanceCount: 10
 
 openTelemetry:
   enabled: true


template/express-rest-api/.gantry/dev.yml
```diff
@@ -1,8 +1,8 @@
-deployment: dev
+deployment: <%- devGantryEnvironmentName %>
 env: development
 
-maxInstanceCount: 1
 minInstanceCount: 1
+maxInstanceCount: 1
```
trevorgk: Swap these two values around. It makes more sense, to me; to read `min` first and then `max`.
Also it caught a team out during a production incident 
https://github.com/SEEK-Jobs/hirer-access-api/commit/b45f858182d86dbde3a1315f5db5edadb59c719f

PR #2124, opened 2025-10-22
Migrate to ESM

samchungy:
Main Changes:

- Migrated to publish skuba as ESM only
- Migrated prettier config to be ESM
- Migrated to vitest
- Deleted Jest modules including GitHub annotations module
- Replace Jest with VItest in pnpm-workspace hoist logic
- Replaced read-pkg-up with read-package-up (I couldn't mock it due to it being a CJS module)

This PR pretty much just gets skuba to work but doesn't do any of the work to migrate. Due to the sheer size of this PR I would like to merge this in first and bring in the patches later

This means we'd be keeping skuba in a not ready to ship state for a period of time.

Future autofixes:
Migrating to Vitest + VItest Presets work will be done here: https://github.com/seek-oss/skuba/pull/2274
Replacing the globals work will be done here: https://github.com/seek-oss/skuba/pull/2139

### DIFF
--- a/.changeset/lovely-islands-argue.md
+++ b/.changeset/lovely-islands-argue.md
@@ -1,5 +1,0 @@
----
-'pnpm-plugin-skuba': major
----
-
-Replace hoisted Jest dependencies with Vitest


template/greeter/src/app.test.ts
```diff
@@ -1,5 +1,7 @@
+import { describe, expect, it } from 'vitest';
+
 import * as app from './app.js';
 
 describe('app', () => {
-  it('is importable', () => expect(app).toEqual({}));
+  it('is importable', () => expect(app).toBeDefined());
```
samchungy: This was failing with:

>AssertionError: expected { [Symbol(Symbol.toStringTag)]: 'Mod…' } to deeply equal {}

src/cli/configure/types.ts
```diff
@@ -1,15 +1,28 @@
...
 export type { TsConfigJson } from 'type-fest';
 
-export type PackageJson = TypeFestPackageJson & Record<string, unknown>;
+export type PackageJson = ReadPackageUpPackageJson &
+  Record<string, unknown> & {
+    skuba?: {
+      entryPoint?: string | null;
+      template?: string;
+      type?: string;
+      version: string;
```
zbrydon: Do we care about defining the `build` field here or not really because it's not generally part of the config setup?

PR #2129, opened 2025-10-30
Apply Vitest/ESM compatibility tweaks

samchungy:
Attempt to make you hate me less for when the big boi drops

### DIFF
--- a/src/cli/lint/internal.test.ts
+++ b/src/cli/lint/internal.test.ts
@@ -2,9 +2,9 @@ import { internalLint } from './internal.js';
 
 describe('internalLint', () => {
   beforeEach(() => {
-    // jest.spyOn(console, 'log').mockImplementation(() => {
-    //   /* no-op */
-    // });
+    jest.spyOn(console, 'log').mockImplementation(() => {
+      /* no-op */
+    });
   });
 
   afterEach(() => {


src/utils/sleep.ts
```diff
@@ -0,0 +1,12 @@
+interface Timeout extends PromiseLike<void> {
+  clear?: () => void;
+}
+
+export const sleep = (ms: number): Timeout => {
```
samchungy: Vitest doesn't support spying on a function which is being tested against within the same file.

eg. we're testing `throwOnTimeout` while expecting `sleep` to be spied on.

src/cli/build/esbuild.ts
```diff
@@ -1,6 +1,6 @@
 import { inspect } from 'util';
 
-import tsconfigPaths from '@esbuild-plugins/tsconfig-paths';
+import { TsconfigPathsPlugin } from '@esbuild-plugins/tsconfig-paths';
```
samchungy: more esm things

src/cli/lint/internal.test.ts
```diff
@@ -2,9 +2,9 @@ import { internalLint } from './internal.js';
 
 describe('internalLint', () => {
   beforeEach(() => {
-    jest.spyOn(console, 'log').mockImplementation(() => {
-      /* no-op */
-    });
+    // jest.spyOn(console, 'log').mockImplementation(() => {
+    //   /* no-op */
+    // });
```
zbrydon: 🤔

zbrydon (approved):

src/utils/fs.ts
samchungy: Initially had this function within `dir.ts` but that imports `@skuba-lib/api` which then creates a circular dependency

PR #2132, opened 2025-10-31
Extract more ESM changes

samchungy:
Rip out more changes from the ESM migration PR

### DIFF
--- a/.changeset/spicy-suns-buy.md
+++ b/.changeset/spicy-suns-buy.md
@@ -1,9 +1,9 @@
 ---
-'skuba': minor
+'skuba': patch
 ---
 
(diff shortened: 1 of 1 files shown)


src/wrapper/http.ts
```diff
@@ -64,21 +63,3 @@ export const serveRequestListener = (
   const server = http.createServer(requestListener);
   return startServer(server, port);
 };
-
-/**
```
samchungy: Once again you can't `spyOn` a function within the same file as something you are testing against

.changeset/spicy-suns-buy.md
samchungy: Changeset just in case we need to send out a release before the migration is complete

PR #2139, opened 2025-11-17
Replace global vars

zhua633:
Adds these patches:

- Replace global vars `__dirname` and`__filename` with `import.meta.dirname` and `import.meta.filename`
- Add `"type": "module"` to `package.json`
- Config files ESM import/export migration
- Add `eslint.config.js` default export rule bypass for config files
- Migrating from `export =` to `export default`


src/cli/migrate/esm/addTypeModuleToPackageJson.test.ts
samchungy, line 1: Can you move everything to `15.3.0`. We use the current version

src/cli/migrate/esm/migrateExportEqualsToDefault.ts
```diff
@@ -0,0 +1,109 @@
...
+    return [];
+  }
+
+  return [match.replace(`export default ${expr.text()};`)];
+};
+
+const resolveParseLanguage = (
+  relativePath: string,
+): 'JavaScript' | 'TypeScript' =>
+  /\.(ts|tsx|mts|cts)$/i.test(relativePath) ? 'TypeScript' : 'JavaScript';
```
samchungy: All your files are in TypeScript, I don't think this is needed

src/cli/migrate/esm/migrateEslintConfigExportDefault.ts
```diff
@@ -0,0 +1,451 @@
...
+    }
+  }
+
+  return edits;
+};
+
+// Adds `with { type: "json" }` to existing default `import … from '…json'` (ESM import attributes).
+const transformExistingJsonDefaultImport = (ast: SgNode): Edit[] => {
+  const imports = ast.findAll({
+    rule: { pattern: 'import $NAME from $SPEC' },
```
samchungy: I think we can simplify this with a better rule:

```yaml
rule:
  kind: 'import_statement'
  has:
    kind: 'string'
    regex: ".json'$"
  not:
    has:
      kind: 'import_attribute'
```

src/cli/migrate/esm/migrateExportEqualsToDefault.ts
```diff
@@ -0,0 +1,109 @@
...
+import fg from 'fast-glob';
+import fs from 'fs-extra';
+
+import { log } from '../../../utils/logging.js';
+import type { PatchFunction } from '../../lint/internalLints/upgrade/index.js';
+import { fetchFiles } from '../../lint/internalLints/upgrade/patches/12.4.1/rewriteSrcImports.js';
+
+const transformExportEqualsToDefault = (ast: SgNode): Edit[] => {
+  const match = ast.find({
+    rule: { pattern: 'export = $EXPR' },
```
samchungy:
```suggestion
    rule: { pattern: 'export = $EXPR', kind: 'export_statement' },
```

src/cli/migrate/esm/migrateImportExportStatements.ts
samchungy, line 1: Does this file name represent what this is doing properly?

src/cli/migrate/esm/migrateImportExportStatements.ts
```diff
@@ -0,0 +1,451 @@
...
+      modulePath,
+      isJsonFile: isJsonModuleSpecifier(modulePath),
+    }),
+  );
+};
+
+// Converts `const x = require('some-module')` to `import x from 'some-module'`
+const transformRequireDeclarations = (ast: SgNode): Edit[] => {
+  const matches = ast.findAll({
+    rule: {
```
samchungy, line 76:
```suggestion
    rule: {
      kind: 'lexical_declaration',
      inside: {
        kind: 'program'
      },
```

Should this be restricted to only top level requires? I think we may need to handle in file `require's` a little differently.

eg. transform them to `const bla = await import('./foo')`

src/cli/migrate/esm/addTypeModuleToPackageJson.ts
```diff
@@ -0,0 +1,119 @@
...
+  PatchReturnType,
+} from '../../lint/internalLints/upgrade/index.js';
+
+export const hasDirNameRegex = /__dirname\b/;
+export const hasFileNameRegex = /__filename\b/;
+
+const PACKAGES_EXCLUDED_FROM_TYPE_MODULE = ['eslint-config-skuba'];
+
+export const addTypeModule = async (cwd: string, originalContent: string) => {
+  const manifest = await readPackageUp({ cwd, normalize: false });
```
samchungy: We're already reading the file, do we want to be finding the package.json and reading again?

src/cli/migrate/esm/rewriteGlobalVars.ts
```diff
@@ -0,0 +1,91 @@
...
+import { inspect } from 'util';
+
+import fg from 'fast-glob';
+import fs from 'fs-extra';
+
+import { log } from '../../../utils/logging.js';
+import type { PatchFunction } from '../../lint/internalLints/upgrade/index.js';
+import { fetchFiles } from '../../lint/internalLints/upgrade/patches/12.4.1/rewriteSrcImports.js';
+
+export const hasDirNameRegex = /__dirname\b/;
```
samchungy: We want to globally replace these, not just one offs

src/cli/migrate/esm/migrateImportExportStatements.ts
zhua633: updated to `migrateImportExportStatements..ts`

src/cli/migrate/esm/addTypeModuleToPackageJson.ts
zhua633: aha good one, nope changed to `parsePackage(originalContent)` instead of reading again

src/cli/migrate/esm/migrateEslintConfigExportDefault.ts
zhua633: thanks! Just got `regex: String.raw`\.json['"]$`` for both single/double quotes matching

src/cli/migrate/esm/rewriteGlobalVars.ts
zhua633: nice one, added /g flag

src/cli/migrate/esm/migrateImportExportStatements.ts
zhua633: ah yeah good one non-top levels gotta use awaits, thanks changed to this:

```
      pattern: 'const $NAME = require($MOD)',
      inside: { kind: 'program' },
```

samchungy (approved):

PR #2142, opened 2025-11-25
Tweak lambda template

samchungy:
- Adds `@aws-sdk/client-secrets-manager` as a dep.
- Adds `datadogApiKeySecretArn` to part of the init

### DIFF
--- a/template/lambda-sqs-worker-cdk/skuba.template.js
+++ b/template/lambda-sqs-worker-cdk/skuba.template.js
@@ -26,13 +26,13 @@ module.exports = {
       name: 'devDataDogApiKeySecretArn',
       message: 'Dev DataDog API Key Secret ARN',
       initial:
-        'arn:aws:secretsmanager:<Region>:<AccountId>:secret:<SecretName>',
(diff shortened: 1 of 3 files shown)


template/lambda-sqs-worker-cdk/skuba.template.js
```diff
@@ -22,6 +22,18 @@ module.exports = {
       initial: 'my-team-aws-account-prod:cicd',
       validate: (value) => /^.+:.+$/.test(value),
     },
+    {
+      name: 'devDataDogApiKeySecretArn',
+      message: 'Dev DataDog API Key Secret ARN',
+      initial:
+        'arn:aws:secretsmanager:<Region>:<AccountId>:secret:<SecretName>',
```
72636c: 1. Is this typically a different ARN between dev and prod?
2. Are our accounts provisioned with a standard secret name by default, and if so, could we use that as an actual default here?

samchungy: 1. I think most teams have a different ARN as they haven't facilitated the sharing mechanism between accounts
2. Looks like `datadog-api-key` is pretty standard

PR #2318, opened 2026-03-30
Release detect-invalid-spies

samchungy:
Vibe coded a little something something for detecting invalid spy usage when migrating from Jest -> Vitest. This was a little difficult to throw into a lint rule and I figure it'll mostly be for 1 time runs to get from Jest -> Vitest.

### DIFF
--- a/packages/detect-invalid-spies/README.md
+++ b/packages/detect-invalid-spies/README.md
@@ -1,0 +1,47 @@
+# @skuba-lib/detect-invalid-spies
+
+Detects `jest.spyOn` / `vi.spyOn` usage patterns where the spy won't work because the spied function is called internally within the same module it is exported from.
+
+## The problem
+
+When you spy on a named export, the spy replaces the binding on the **module namespace object**. But if the source module calls that function directly (using its local binding), those calls bypass the spy entirely.
+
+```ts
+// greet.ts
+export const sayHello = (name: string) => `Hello, ${name}!`;
+
+export const greetEveryone = (names: string[]) =>
+  names.map((name) => sayHello(name));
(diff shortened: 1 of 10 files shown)


packages/detect-invalid-spies/src/index.ts
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  const ast = await parseAsync('TypeScript', content);
+  const root = ast.root();
+
+  // Find every identifier whose text exactly matches the function name.
+  const identifiers = root.findAll({
+    rule: {
+      kind: 'identifier',
+      regex: `^${functionName}$`,
```
zbrydon, line 253: Can/should we be bothered to escape this just in case people have some funky function names?

packages/detect-invalid-spies/README.md
zbrydon, line 1: Should we call out that this won't detect re exporting things eg. something weird like this

```ts
// impl.ts
export const doThing = () => 'x';
export const caller = () => doThing();
```

```ts
// index.ts
export { doThing as default } from './impl';
```

```ts
// spy.ts
import * as index from './index.ts';

jest.spyOn(index, 'default');
```

zbrydon (approved):

packages/detect-invalid-spies/src/index.ts
samchungy: Nah it should be okay, we aren't automatically running this via lint or anything so nothing malicious should result of it

packages/detect-invalid-spies/README.md
samchungy: I reckon that's a skill issue if so

PR #2324, opened 2026-04-08
Support named `app` export

samchungy:
Was investigating migration paths for `export =` syntax which a number of repos are currently using.

We're going with the simple `export default` alternative but we should probably support a named export too. `default` exports are generally problematic


zbrydon (approved):

src/wrapper/requestListener.ts
```diff
@@ -54,6 +55,11 @@ export const runRequestListener = async ({
     config = await config.default;
   }
 
+  if (typeof config === 'object' && isConfig(config.app)) {
+    // named export support, e.g. `export const app = new Koa()`
+    config = await config.app;
```
72636c, line 55: Are we comfortable that `export const app` takes precedence over `export default`?

Seems reasonable enough to me, but just checking, and potentially worth documenting.

72636c (approved):
👏

src/wrapper/requestListener.ts
samchungy: Hmm, I might actually go default over app

PR #2326, opened 2026-04-10
Update service templates to prune dev deps when building prod image

kosanna:
Currently, CVE scanning across the board picks up dev dependencies being include in prod images. This PR changes Dockerfiles to use `pnpm prune` instead of just re-installing prod-only deps (which does not prune anything).

https://pnpm.io/cli/prune#--prod

### DIFF
--- a/.changeset/jolly-sloths-argue.md
+++ b/.changeset/jolly-sloths-argue.md
@@ -1,0 +1,7 @@
+---
+'skuba': minor
+---
+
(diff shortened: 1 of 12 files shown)


72636c (approved):
This looks ace, thank you.

For posterity, we've been observing inconsistent `pnpm install --prod` behaviour in different environments, even when running `pnpm install --prod` without a preceding `pnpm install`, meaning that having a separate stage won't fix the issue. `pnpm prune --prod` appears to consistently work in comparison, so we figure this is the best way forward for now.

PR #2342, opened 2026-04-24
Fix `skuba start` watch

zbrydon:


samchungy:
Interestingly this looks like it works while in ESM 🤷

zbrydon:
> Interestingly this looks like it works while in ESM 🤷

Given that this would only be released with ESM may as well close this?

zbrydon:
This works without changes once we're on esm so closing this.

PR #2343, opened 2026-04-26
Unbundle patch

zbrydon:


samchungy:
Legend, we might park this til after ESM though to try and reduce the number of things going on

PR #2355, opened 2026-05-06
Patch instrumentation libraries

samchungy:
https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/nodejs/#esm-applications-only-import-the-loader

and

https://github.com/open-telemetry/opentelemetry-js/blob/main/doc/esm-support.md

### DIFF
--- a/src/cli/migrate/esm/patchInstrumentation.ts
+++ b/src/cli/migrate/esm/patchInstrumentation.ts
@@ -69,6 +69,11 @@ export const patchInstrumentation: PatchFunction = async ({
     );
   }
 
+  const warning =
+    hasDDTraceImport && hasOpenTelemetryImport
+      ? 'TODO: skuba failed to determine whether to add dd-trace or OpenTelemetry flags, please choose the appropriate flags to add to your Dockerfile '
+      : '';
+
   const commandsToAdd = [
     hasDDTraceImport ? '--import dd-trace/initialize.mjs' : '',
     hasOpenTelemetryImport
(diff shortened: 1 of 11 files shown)


src/cli/migrate/esm/patchInstrumentation.ts
```diff
@@ -0,0 +1,203 @@
...
+    };
+  }
+
+  const confused = hasDDTraceImport && hasOpenTelemetryImport;
+
+  if (confused) {
+    log.warn(
+      'Found imports for both Datadog and OpenTelemetry instrumentation in source files, unsure which to patch',
+    );
+  }
```
zbrydon, line 70: Thinking we just log and try both over log and do nothing?

src/cli/migrate/esm/patchInstrumentation.ts
```diff
@@ -0,0 +1,203 @@
...
+          const patchedCmd = cmd.replace(
+            'node ',
+            `node ${commandsToAdd.join(' ')} `,
+          );
+          return { filePath, content: content.replace(cmd, patchedCmd) };
+        }
+
+        return {
+          filePath,
+          content: content.replace(cmd, `${commandsToAdd.join(' ')} ${cmd}`),
```
zbrydon, line 110: Should this be adding the commands if there's no `node`?

zbrydon (approved):

src/cli/migrate/esm/patchInstrumentation.ts
samchungy: Decided to just straight up break the Dockerfile if there's both so the consumer will have to manually resolve: [`4ac4541` (this PR)](https://github.com/seek-oss/skuba/pull/2355/commits/4ac4541d120e6e0c63a109ed2f0c922dd414748d)

src/cli/migrate/esm/patchInstrumentation.ts
samchungy: Yeah, the command implicitly invokes `node` by default: https://github.com/seek-oss/skuba/blob/e3a0d1d1d28f3770a797a829290f259b9332312f/template/koa-rest-api/Dockerfile#L32

PR #2387, opened 2026-05-11
Add SEEK registry to .npmrc

### LINKED ISSUE #2378: Autofix .npmrc into repos
https://seekchat.slack.com/archives/C06QWMV5BSA/p1776390927664839

.npmrc is back y'all

zbrydon:
Resolves https://github.com/seek-oss/skuba/issues/2378

---

I wonder if we should use the `pnpm-workspace.yaml` instead 🤔 
https://github.com/pnpm/pnpm/releases/tag/v11.1.0

---

I tried out the above and turns out it's implemented on the version specifier not the package name, eg. 
```json
// package.json
"@seek/some-package": "seek:1.0.0",
```
```yml
// pnpm-workspace.yaml
registries:
  seek: https://our-special-seek-registry.com
```

As a result if we were to remove the equivalent entry from the `.npmrc` file it would require every transitive dependency that depends on a `@seek` scoped privately hosted package to also use the `seek:` prefix in their version. 

This seems like way to much effort for just avoiding having to add a repository level `.npmrc` file just because it's slightly easier to accidentally leak tokens.


src/cli/lint/internalLints/upgrade/patches/16.1.0/addSeekPackageRegistry.ts
```diff
@@ -0,0 +1,84 @@
...
+  const remotes = await git.listRemotes({ fs, dir: process.cwd() });
+  const originUrl = remotes.find((r) => r.remote === 'origin')?.url;
+  const org = originUrl?.match(/github\.com[:/]([^/]+)\//)?.[1];
+
+  if (org !== SEEK_ORG) {
+    return {
+      result: 'skip',
+      reason: 'not a SEEK-Jobs repository',
+    };
+  }
```
zbrydon: Thoughts on adding it selectively?

samchungy: Could, we generally use https://github.com/seek-oss/skuba/blob/cefe2f57042562eff0d85ef0557c0096b025db75/packages/api/src/git/remote.ts#L47

src/cli/lint/internalLints/upgrade/patches/16.1.0/addSeekPackageRegistry.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  if (npmrcPaths.length === 0) {
+    return {
+      result: 'skip',
+      reason: 'no .npmrc files found',
+    };
+  }
+
+  const npmrcFiles = await Promise.all(
```
samchungy, line 52: I think we need to write a .npmrc for repos without one?

zbrydon: 😆 ofc 🤦

src/cli/lint/internalLints/upgrade/patches/16.1.0/addSeekPackageRegistry.ts
zbrydon: Ah yes nice TIL 🙏

src/cli/lint/internalLints/upgrade/patches/16.1.0/addSeekPackageRegistry.ts
samchungy: Probably want to go with:

Finding the git root and writing it there, as the resolution would only happen at the root generally 🤔

src/cli/lint/internalLints/refreshConfigFiles.ts
```diff
@@ -24,39 +25,36 @@ type ConditionOptions = {
...
-
 export const REFRESHABLE_CONFIG_FILES: RefreshableConfigFile[] = [
   {
     name: '.gitignore',
     type: 'ignore',
-    additionalMapping: removeOldWarning,
   },
   { name: '.prettierignore', type: 'ignore' },
+  {
+    name: '.npmrc',
```
samchungy, line 51: do we need this? Isn't this only for files we have in template/base?

zbrydon: mhmmm true, the intention was to be ~extra~ slightly more sure that secrets don't end up in it 😅 but not sure we want to make this part of the base templates 🫠

samchungy (approved):

PR #2396, opened 2026-05-12
Update file extensions patch to handle adding type imports extensions

samchungy:
Missed a case


src/cli/migrate/esm/vitest/vitestConfig.ts
```diff
@@ -669,16 +669,16 @@ const scaffoldTestConfig = async ({
         functions: 100,
         lines: 100,
         statements: 100,
-      }${
-        !isProject && (configNumber ?? 0) > 1
```
samchungy, line 673: Accidentally stashed this TODO into the coverage

src/cli/migrate/esm/addFileExtensions.ts
```diff
@@ -151,19 +161,13 @@ export const addFileExtensions = async ({
 
       const parentPath = pathToFileURL(file).href;
       const resolvedPaths = new Map<string, string | null>();
```
samchungy, line 163: Since the edits were in a Promise.all this wasn't getting cached properly so I've rewritten this to use it properly

zbrydon (approved):

PR #2406, opened 2026-05-13
Fix spy detection

samchungy:
Handles spies which are broken up by prettier


packages/detect-invalid-spies/package.json
```diff
@@ -49,6 +49,9 @@
     "node": ">=22.14.0"
   },
   "publishConfig": {
+    "bin": {
```
samchungy, line 52: tsdown insisted

zbrydon (approved):

PR #2412, opened 2026-05-13
Adjust import types patch insertion logic

samchungy:
Ugh

import-x has this kinda buggy behavior if you have duplicate imports in your code which aren't sorted behind the correct imports where it nukes any `// eslint-disable-next-line import-x/order` line causing issues for our migrations :( because the `'#src/mockLogger.js'` then gets sorted amongst the regular imports and the logger is left unmocked

eg.
```ts
// eslint-disable-next-line import-x/order -- contains mocks
import {testLogger} from '#src/mockLogger.js';

import { vi } from 'vitest';
import thing from 'thing';

import { logger } from '#src/logger.js';
import type { Mock } from 'vitest';
```

But if you moved `import type { Mock } from 'vitest';` above any `#src/` imports it's fine 🤷 

Running format completely nukes the `// eslint-disable-next-line import-x/order` comment and gets lost in the weeds for our consumers migrating. This causes pain because you only realize this was done after the tests fail.

So the workaround I've got going is just double parse the AST and attempt to insert the `vitest` line directly under the existing vitest import so we don't have issues.

### DIFF
--- a/src/cli/migrate/esm/vitest/postFixVitestMigration.ts
+++ b/src/cli/migrate/esm/vitest/postFixVitestMigration.ts
@@ -293,7 +293,19 @@ const getTypeImportEdits = (root: SgNode, imports: string[]): Edit[] => {
     return [];
   }
 
-  const lastImport =
+  const importPlacement =
+    root.find({
+      rule: {
+        kind: 'import_statement',
+        has: {
+          kind: 'string',
+          has: {
+            kind: 'string_fragment',
+            regex: '^vitest$',
+          },
+        },
+      },
+    }) ??
     root.find({
       rule: {
         kind: 'import_statement',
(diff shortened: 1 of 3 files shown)


src/cli/migrate/esm/vitest/postFixVitestMigration.ts
```diff
@@ -293,7 +293,48 @@ const getTypeImportEdits = (root: SgNode, imports: string[]): Edit[] => {
...
+          kind: 'string',
+          has: {
+            kind: 'string_fragment',
+            regex: '^vitest$',
+          },
+        },
+      },
+    }) ??
+    root.find({
+      rule: {
```
samchungy, line 310: fallback to after the last lettered import

PR #2417, opened 2026-05-15
Remove `skuba configure`

### LINKED ISSUE #1120: Drop ancient `configure` logic
There are a number of configuration modules that we can probably drop in {{currentYear}}. For example, we have code to delete TSLint config files:

https://github.com/seek-oss/skuba/blob/dbd485355223fd914169bbfa4c4ae792eaa6beb9/src/cli/configure/modules/tslint.ts#L4-L5

zbrydon:
Resolves https://github.com/seek-oss/skuba/issues/1120


samchungy (approved):

72636c:
Apologies, scope change in #1120 🥲

docs/templates/byo.md
```diff
@@ -15,6 +15,4 @@ This allows you to build and maintain template repositories with your own prefer
 ## github
 
 **skuba** will shallow-clone your template repo from GitHub and apply some of its base configuration on top.
-You may need to run [`skuba configure`] after initialisation to fix up inconsistencies.
-
-[`skuba configure`]: ../cli/configure.md#skuba-configure
+You may need to manually complete templating after initialisation to fix up inconsistencies.
```
72636c:
```suggestion
```

src/cli/init/getConfig.ts
```diff
@@ -138,13 +135,7 @@ const cloneTemplate = async (
...
     log.newline();
-    log.warn(
-      'You may need to run',
-      log.bold(
-        `${configForPackageManager(templateConfig.packageManager).print.exec} skuba configure`,
-      ),
-      'once this is done.',
-    );
+    log.warn('You may need to manually complete templating once this is done.');
   }
```
72636c: I don't think it's necessarily "templating" that needs to be fixed up at this point, it's more if the custom template starts with non-standard stuff that `skuba configure` would then overwrite.

I think if you BYO template it's reasonable to assume that you'd be okay with starting with the conventions of the template 🤷

```suggestion
```

src/cli/init/getConfig.ts
```diff
@@ -293,9 +284,7 @@ export const configureFromPrompt = async (): Promise<InitConfig> => {
...
 
   log.newline();
-  log.warn(
-    `Resume this later with ${styleText('bold', `${configForPackageManager(packageManager).print.exec} skuba configure`)}.`,
-  );
+  log.warn('Templating has been skipped. Complete it manually before running.');
```
72636c: This part is actually unfortunate, I forgot that `skuba configure` manages this "resume templating" bit. Thoughts on options here?

1. Remove the option to skip templating
2. Keep subsequent templating manual (current state of this PR)
3. Retain very rudimentary command for this purpose (not the catch-all nature of the old `configure`)

72636c (approved):

src/cli/init/getConfig.ts
zbrydon: IIRC the optional templating section has some fairly SEEK specific things in it, which kinda makes sense but then there's also an argument for not making external consumers go through the templating just to delete it again 😆 

I personally don't think completing it manually is that onerous ie. it's really just a find and replace but I can see that causing headaches for some 🙈

samchungy:
changeset?

src/cli/init/getConfig.ts
zbrydon: Agreed path forward is to make `skuba init` pick up from where it left off when run again for the optional templating section.

samchungy (approved):

72636c (approved):

Issue #2035, opened 2025-08-19
Consider hiding cdk template redaction regex


zbrydon:
It has been raised that 
https://github.com/seek-oss/skuba/blob/343795b54de15c5f38538a4a72589958728b7faa/template/lambda-sqs-worker-cdk/infra/appStack.test.ts#L40-L58

is a bit of an eye sore. 

Aaron suggested that we could hide some of the ~atrocities~ common issues in an internal package.

zbrydon:
https://github.com/seek-oss/skuba/pull/2418

PR #2418, opened 2026-05-16
Move CDK normalisation logic to `@skuba-lib/api`

zbrydon:
https://github.com/seek-oss/skuba/issues/2035


samchungy:
I don't think we wanna stash it in skuba dive, as it's a runtime dep. Wouldn't be opposed to stashing it in the skuba lib api and re-export via skuba

zbrydon:
> I don't think we wanna stash it in skuba dive, as it's a runtime dep. Wouldn't be opposed to stashing it in the skuba lib api and re-export via skuba

Yeah fair enough I didn't like it either but started as a standalone package and that felt worse 😅

.changeset/khaki-rivers-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'skuba': minor
```
samchungy:
```suggestion
'skuba': patch
```

.changeset/fine-jars-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+'@skuba-lib/api': minor
```
samchungy, line 2:
```suggestion
'@skuba-lib/api': minor
'skuba': minor
```

samchungy (approved):

PR #2430, opened 2026-05-20
Allow re-runs of the Vitest migration

samchungy:
Came up as a bit of a pain point of needing to pull changes in from other people's merges in Hirer Graph

### DIFF
--- a/src/cli/migrate/esm/vitest/vitest.ts
+++ b/src/cli/migrate/esm/vitest/vitest.ts
@@ -140,10 +140,14 @@ export const migrateToVitest = async (opts: {
     ignore: ['**/.git', '**/node_modules'],
   });
 
-  if (vitestConfigFiles.length > 0) {
+  if (
+    vitestConfigFiles.length > 0 &&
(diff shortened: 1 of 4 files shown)


src/cli/migrate/esm/vitest/vitest.ts
```diff
@@ -265,7 +269,7 @@ export const migrateToVitest = async (opts: {
         )
         .replaceAll('eslint-disable jest', 'eslint-disable vitest')
         .replaceAll('Mock<any, any>', 'Mock')
-        .replaceAll('advanceTimers: true', 'shouldAdvanceTimers: true');
+        .replaceAll('advanceTimers: true', 'shouldAdvanceTime: true');
```
samchungy, line 272: Little mistake

zbrydon (approved):
It's applying to a different command but quite similar to this 🤔 
https://github.com/seek-oss/skuba/blob/de6422a944d1a0cba4050ddec01c8b64dd36081d/docs/cli/lint.md?plain=1#L84-L85

PR #2435, opened 2026-05-27
Update pnpm-workspace processing logic

samchungy:
> tl;dr: I was a noob and didn't know how to use `@ast-grep` effectively and have made the patch better.

Bit of a hack until https://github.com/arethetypeswrong/arethetypeswrong.github.io/issues/258 is resolved to unblock our repos

This was initially a hack to quickly resolve our issues but I've re-written it completely to handle more scenarios.

Basically this whole script does:

1. Look for lines with `# Managed by skuba` to delete which don't match our config.
2. Delete any sections which may be left orphaned after deleting.
3. Add missing managed sections
4. Update existing sections.
5. Run `pnpm install` if we've added an override to update the lockfile

### DIFF
--- a/src/cli/lint/internalLints/patchPnpmWorkspace.test.ts
+++ b/src/cli/lint/internalLints/patchPnpmWorkspace.test.ts
@@ -84,6 +84,8 @@ describe('patchPnpmWorkspace', () => {
       trustPolicy: off # Managed by skuba
       trustPolicyExclude:
         - semver@6.3.1 # Managed by skuba
+      overrides:
+        '@arethetypeswrong/core@0.18.2>fflate': 0.8.2 # Managed by skuba
       "
     `);
   });
@@ -159,6 +161,8 @@ trustPolicyExclude:
       packageManagerStrictVersion: true # Managed by skuba
       strictDepBuilds: false # Managed by skuba
       trustPolicy: off # Managed by skuba
+      overrides:
+        '@arethetypeswrong/core@0.18.2>fflate': 0.8.2 # Managed by skuba
       # This is a comment
       allowBuilds:
         '@datadog/native-appsec': true # Managed by skuba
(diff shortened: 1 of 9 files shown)


tsconfig.json
```diff
@@ -3,10 +3,10 @@
     "rootDir": ".",
     "customConditions": ["@seek/skuba/source"],
     "declaration": true,
-    "lib": ["ES2023"],
+    "lib": ["ES2024"],
```
samchungy, line 6: Should be non breaking unless we use Unicode 16/17 in RegExp v flags

src/cli/lint/internalLints/patchPnpmWorkspace.ts
```diff
@@ -344,14 +585,71 @@ export const patchPnpmWorkspace = async (
...
-
   await fs.promises.writeFile(
     path.join(dir, 'pnpm-workspace.yaml'),
-    newSource,
+    finalSource,
     'utf8',
   );
 
+  const finalAst = (await parseAsync('yaml', finalSource)).root();
+  const lockFileUpdateTriggers = ['overrides'];
```
zbrydon: Would this not extend to most settings? eg we bump the minimum release age or add an allowed build and want to check that the content of the lock file is compliant?

src/cli/lint/internalLints/patchPnpmWorkspace.test.ts
zbrydon: Should this line get removed? 🤔

zbrydon (approved):

src/cli/lint/internalLints/patchPnpmWorkspace.test.ts
samchungy: Stray comment? Hmmm sure we can nuke em

src/cli/lint/internalLints/patchPnpmWorkspace.ts
samchungy: image

🤔 hmmm I need to sus which ones trigger

samchungy: okay so none of the minimum release age updates or allowed builds trigger lockfile updates, just the overrides

PR #2444, opened 2026-06-02
Fix esm lambda apm

zbrydon:
I've created an issue in `datadog-lambda-js` around this and have a support ticket open with them. 

Should we wait for them to get back to me or proceed with attempting to work around it for now then revert things if a fix is released?


samchungy (approved):

src/cli/lint/internalLints/upgrade/patches/16.1.0/addDdTraceEsmImport.ts
```diff
@@ -0,0 +1,177 @@
...
+      kind: 'new_expression',
+      regex: '^new (aws_lambda_nodejs.NodejsFunction|NodejsFunction)',
+    },
+  });
+
+  if (!workerAst) {
+    return null;
+  }
+
+  const nodeModulesText =
```
samchungy: I feel like if they have redirectHandlers we can just do the patch tbh

samchungy (approved):

PR #2447, opened 2026-06-03
Remove @arethetypeswrong/core override

samchungy:
They've fixed the thing


packages/detect-invalid-spies/src/bin.ts
samchungy, line 1: Don't fully get what's going on here but when running install this gets changed, guess making it executable is fine given it's a ts file anyway

zbrydon (approved):

PR #2443, opened 2026-06-02
Add multi stage builds

zbrydon:

PR #2451, opened 2026-06-05
Rework Dockerfile patch

zbrydon:
Sam mentioned #2443 won't work with some common monorepo setups so opting for a simple default

### DIFF
--- a/.changeset/clear-terms-shop.md
+++ b/.changeset/clear-terms-shop.md
@@ -10,8 +10,6 @@ the build stage before building. The runtime image continues to copy
 development dependencies carried through to the final image.
 
 ```diff
-ARG BASE_IMAGE
-
 FROM ${BASE_IMAGE} AS build
 COPY . .
 
@@ -19,13 +17,4 @@ COPY . .
 + RUN pnpm prune --prod
 + RUN pnpm install --offline --prod
 RUN pnpm build
-
-###
-
-FROM gcr.io/distroless/nodejs24-debian13 AS runtime
-WORKDIR /workdir
-
-COPY --from=build /workdir/lib lib
-COPY --from=build /workdir/package.json package.json
-COPY --from=build /workdir/node_modules node_modules
 ```
--- a/.changeset/clear-terms-shop.md
+++ b/.changeset/clear-terms-shop.md
@@ -2,28 +2,23 @@
 'skuba': minor
 ---
 
-template/\*-rest-api: Add multi-stage Docker builds
+template/\*-rest-api: Prune development dependencies from Docker builds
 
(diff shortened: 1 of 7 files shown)


.changeset/clear-terms-shop.md
```diff
@@ -2,36 +2,19 @@
...
-+ RUN pnpm prune --prod
-+ RUN pnpm install --offline --prod
-+
-###
-
 FROM ${BASE_IMAGE} AS build
 COPY . .
 
-RUN pnpm install --offline
+- RUN pnpm install --offline
```
samchungy: So we still need this one,

Should go

```bash
RUN pnpm install --offline
RUN pnpm build
RUN pnpm prune --prod
RUN pnpm install --offline --prod
```

samchungy: Basically:

Install dev deps, run the build, nuke the dev deps

zbrydon: 🫠 ahhh yep gotcha

.changeset/clear-terms-shop.md
```diff
@@ -2,36 +2,19 @@
...
-###
-
-FROM gcr.io/distroless/nodejs24-debian13 AS runtime
-WORKDIR /workdir
-
-COPY --from=build /workdir/lib lib
-- COPY --from=build /workdir/node_modules node_modules
-COPY --from=build /workdir/package.json package.json
-+ COPY --from=deps /workdir/node_modules node_modules
++ RUN pnpm prune --prod
```
samchungy: I think we might need RUN CI=true here because there may be a prompt if we want to delete the node modules or not that we need to bypass.

src/cli/lint/internalLints/upgrade/patches/16.1.0/pruneDevDeps.ts
```diff
@@ -8,14 +8,11 @@ import { log } from '../../../../../../utils/logging.js';
...
-const applyMultiStageBuildPatch = async (
+const applyPruneDevDepsPatch = async (
   contents: string,
 ): Promise<string | null> => {
   // Fairly lazy check to bail out on the patch
-  if (
-    contents.includes('COPY --from=deps') ||
-    contents.includes('RUN pnpm prune --prod')
-  ) {
+  if (contents.includes('pnpm prune --prod')) {
```
samchungy, line 15: I think we should also skip if they are also using pnpm deploy

.changeset/clear-terms-shop.md
zbrydon: Ahhh true that env var is not necessarily passed into the docker image

samchungy (approved):

PR #2458, opened 2026-06-15
Add `Cdk.NodejsFunction`

zbrydon:
### Motivation

The main driver is making the CDK `NodejsFunction` construct work with the strict pnpm settings we want to enforce, and those enforced by default in pnpm 11. The upstream construct doesn't play nicely with these, and thought we can use rolldown instead of esbuild while we're at it.

Related upstream issues:
- aws-cdk: https://github.com/aws/aws-cdk/issues/37898
- pnpm: https://github.com/pnpm/pnpm/issues/10988

### Overview

A `NodejsFunction` CDK construct under `@skuba-lib/api/cdk`, a drop-in alternative to `aws-cdk-lib/aws-lambda-nodejs.NodejsFunction`, built around **rolldown** and **pnpm** . At synth time it bundles a TS/JS entry into a single ESM `index.mjs` handler, optionally installs a subset of real `node_modules`, and packages the result as a Lambda asset.

Flow: `Cdk.NodejsFunction` → `function.ts` (validation) → `bundling.ts` (`tryBundle`) → spawns the rolldown bridge → optional `pnpm install` → `outputDir` becomes the asset.

### Bits worth noting

- **`cdk/index.ts` is a lazy `require()` getter, not a normal export.** `aws-cdk-lib` and `constructs` are optional peers; the getter defers loading them so consumers who never use the construct don't need them installed. See [discussion](https://github.com/seek-oss/skuba/pull/2458#discussion_r3410484616) on potential options 
- **Local bundling only, no Docker fallback.** `Bundling` implements CDK's `BundlingOptions` but the `image` is a dummy and `tryBundle` always returns `true`. Synth therefore requires `node`, `pnpm`, and rolldown in the environment.
- **rolldown runs out-of-process** via a compiled bridge (`bridges/rolldown.mts`) resolved from the package root. This isolates the bundler's native bits and module resolution from the CDK app process. The bridge also enforces the single-handler contract: it *rejects* `input`, `output.preserveModules`, `output.entryFileNames`, multi-entry `output`, and non-ESM `format` rather than overriding them.
- **pnpm-only by assertion, not just default.** `package-manager.ts` throws if a non-pnpm package manager is declared.
- **`nodeModules` install path has real subtlety:** pins exact versions via `findPackageJSON`, filters `patchedDependencies` down to the requested modules, copies patch files with path-traversal/symlink guards, and cleans staged files in a `finally`.
- **The bridge path `lib/cdk/bridges/rolldown.mjs` couples source layout to build output**
- **ESM only** Sam suggested simplify things and make it esm only and I extended that to the whole `@skuba-lib/api` because it seems odd to have half esm only 🤷 

### Tests
- https://github.com/SEEK-Jobs/indie-metrics-api/pull/1314
- https://github.com/SEEK-Jobs/indie-moat/pull/1061
- https://github.com/SEEK-Jobs/datadog-organization-data/pull/687


config/tsconfig.json
```diff
@@ -8,6 +8,7 @@
     "noUnusedLocals": false,
     "noUnusedParameters": false,
     "resolveJsonModule": true,
+    "skipLibCheck": true,
```
zbrydon, line 11: This is a result of trying to work around the optional peer deps added in this PR. 

I am not sure about which approach we want to go with 🤷 : 
- make the deps non-optional (I've make rolldown a dep but the other two aren't feasible and could result in version conflicts 🤔 ) 
- lazy load and skip type checking (current PR)
- create a dedicated subpath export via skuba (Seeing what that would look like here https://github.com/seek-oss/skuba/pull/2474)
- just import from the skuba-api (Means we have to make all the consumers of this have another dep 🫠 )

packages/api/scripts/fix-types.sh
zbrydon, line 1: Fixed in https://github.com/sxzz/rolldown-plugin-dts/releases/tag/v0.25.1

zbrydon:
Going to investigate a rolldown plugin + `lambda.Code.fromAsset()` so marking this as draft until I get that together

PR #2463, opened 2026-06-15
Don't double install after `pnpm prune`

zbrydon:
Avoids doing this 🙈 
image

### DIFF
--- a/src/cli/lint/internalLints/upgrade/patches/16.1.0/pruneDevDeps.ts
+++ b/src/cli/lint/internalLints/upgrade/patches/16.1.0/pruneDevDeps.ts
@@ -11,12 +11,8 @@ import type { PatchFunction, PatchReturnType } from '../../index.js';
 const applyPruneDevDepsPatch = async (
   contents: string,
 ): Promise<string | null> => {
-  // Fairly lazy check to bail out on the patch
-  if (contents.includes('pnpm prune --prod')) {
-    return null;
-  }
-
   registerAstGrepLanguages();
+
   const astRoot = (await parseAsync('bash', contents)).root();
 
   const argBaseImage = astRoot.find({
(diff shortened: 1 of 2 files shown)


src/cli/lint/internalLints/upgrade/patches/16.1.0/pruneDevDeps.ts
```diff
@@ -67,11 +67,24 @@ const applyPruneDevDepsPatch = async (
...
+
+  const edits: Edit[] = installProd
+    ? [
+        installProd.replace(
+          'RUN CI=true pnpm prune --prod\nRUN CI=true pnpm install --offline --prod',
+        ),
+      ]
+    : [
+        pnpmBuild.replace(
+          'RUN pnpm build\nRUN CI=true pnpm prune --prod\nRUN CI=true pnpm install --offline --prod',
```
samchungy: I don't think I looked at this properly the first time but rather looked at the tests.

Could we simply just avoid caring about `pnpm build` and instead just look for:

the `CI=true pnpm install --offline --prod` line and insert a prune before it?

src/cli/lint/internalLints/upgrade/patches/16.1.0/pruneDevDeps.ts
```diff
@@ -67,11 +67,24 @@ const applyPruneDevDepsPatch = async (
...
 
-  const edits: Edit[] = [
-    pnpmBuild.replace(
-      'RUN pnpm build\nRUN CI=true pnpm prune --prod\nRUN CI=true pnpm install --offline --prod',
-    ),
-  ];
+  const installProd = astRoot.find({
+    rule: {
+      kind: 'command',
+      regex: '^RUN CI=true pnpm install --offline --prod$',
```
72636c: Are we potentially missing out on coverage by being super specific in these rule matches (e.g. are there some Dockerfiles that omit the `CI=true`)?

72636c (approved):

samchungy (approved):

PR #2464, opened 2026-06-15
Tweak more changesets

72636c:


.changeset/wet-rice-tan.md
```diff
@@ -2,6 +2,18 @@
...
+
+```diff
+- import { bar } from 'src/bar.js';
+
++ // Subpath import option
++ import { bar } from '#src/bar.js';
+
++ // Relative import option
++ import { bar } from './bar.js';
+```
```
72636c, line 19: I made this up because "verify that your project does not depend on it" was not very actionable. Please check for accuracy @samchungy

.changeset/kind-lights-mate.md
zbrydon, line 1: What's our approach on whether to add a diff or not? I was contemplating it for this but wasn't sure 🤔

zbrydon (approved):

.changeset/kind-lights-mate.md
72636c: I'd say where

- It's difficult to explain the change with prose
- An engineer may need to manually make the change

This one seems borderline so I have no strong opinion on it

.changeset/wet-rice-tan.md
samchungy: Nah that works

samchungy (approved):

PR #2466, opened 2026-06-16
Add `.npmrc` to templates

zbrydon:
@72636c @samchungy and I are unsure if this is something we want or not. 

Current state any new project will not get the new `.npmrc` file and will not be compliant

Adding it to the templates breaks the integration tests (which is easy to hack around) but means any external consumers would have to remove this file before it is usable, and I think it would break during the install in the init 🤔

### DIFF
--- a/.changeset/flat-spoons-give.md
+++ b/.changeset/flat-spoons-give.md
@@ -1,0 +1,5 @@
+---
+'skuba': patch
+---
+
+template/\*: Sets `@seek:registry` in `.npmrc` to SEEK's private registry in templates typically used for internal projects.
--- a/.changeset/flat-spoons-give.md
+++ b/.changeset/flat-spoons-give.md
@@ -2,4 +2,4 @@
 'skuba': patch
 ---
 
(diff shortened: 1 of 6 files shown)


72636c:
Could this be the impetus to migrate most of these templates to an internal repository? (Perhaps `greeter` can remain in both.)

zbrydon:
> Could this be the impetus to migrate most of these templates to an internal repository? (Perhaps `greeter` can remain in both.)

Yeah could be I agree, would we want to keep say the lambda and open-source package templates here as they have less SEEK things in them 🤔 

Or is the duplication not worth it...

72636c:
`oss-npm-package` definitely. For Lambda, I'm mostly unsure about whether Buildkite is appropriate scaffolding for personal/external projects.

zbrydon:
> `oss-npm-package` definitely. For Lambda, I'm mostly unsure about whether Buildkite is appropriate scaffolding for personal/external projects.

yeah fail call, have gone with these instead
https://github.com/SEEK-Jobs/skuba-templates/pull/26
https://github.com/SEEK-Jobs/skuba-templates/pull/27

And might close this in favour of just removing the templates from here? 🤔

PR #2473, opened 2026-06-17
Strip underscore prefixes from private templates

etaoins:
When testing SEEK-Jobs/skuba-templates#20 locally, everything worked find because e.g. `_package.json` was being renamed to `package.json`.

However, once the PR was merged I was unable to actually use the template because underscore stripping is disabled for `seek` packages. Testing with the existing `private-npm-package` template fails with the same error.


zbrydon (approved):
Whoops 😅 🙈

72636c (approved):

PR #2481, opened 2026-06-18
Try to automerge non-major GitHub actions

72636c:
I don't care enough about e.g. seek-oss/skuba#2477


zbrydon (approved):
We'll figure it out if it turns out a minor is actually a major in disguise anyways 🤷

PR #2486, opened 2026-06-21
Add `mount-buildkite-agent` patch

### LINKED ISSUE #1945: Patch `mount-buildkite-agent`
https://github.com/seek-oss/skuba/pull/1944

zbrydon:
Resolves #1945


samchungy (approved):

.changeset/moody-melons-wink.md
```diff
@@ -0,0 +1,29 @@
+---
+'skuba': patch
+---
+
+lint: Add patch to migrate Docker Compose Buildkite plugins to `mount-buildkite-agent`
```
72636c: Do we want to fix up Docker Buildkite plugin usage too?

https://github.com/SEEK-Jobs/static-site-router/blob/fdd5b7d4800ab9cad081edcd635f7aa1d5a252fa/.buildkite/pipeline.yml#L24-L31

```diff
  steps:
    - plugins:
        - *docker-ecr-cache
        - docker#v5.13.0:
-           # Disable SEEK BuildAgency's wrapped agent that requires Bash.
-           mount-buildkite-agent: false
+           mount-buildkite-agent: true
            propagate-environment: true
            volumes:
-             # Mount agent for Buildkite annotations.
-             - /usr/bin/buildkite-agent:/usr/bin/buildkite-agent
-             # Mount cached dependencies.
              - /workdir/node_modules
```

72636c (approved):
🤩

src/cli/lint/internalLints/upgrade/patches/16.2.0/mountBuildkiteAgent.ts
```diff
@@ -0,0 +1,296 @@
...
+    .filter((child) => child.kind() === 'block_mapping_pair');
+
+  const [firstOption] = options;
+  const lastOption = options.at(-1);
+
+  if (
+    !firstOption ||
+    !lastOption ||
+    options.some(
+      (option) => option.field('key')?.text() === MOUNT_BUILDKITE_AGENT,
```
KSHITIZ6341: I think this may need attention before merge because the Docker Compose path treats any existing `mount-buildkite-agent` key as done, even when it is explicitly `false`.

`addMountBuildkiteAgentEdits` returns no edits as soon as the option exists. If a project currently has both a `docker-compose#...` plugin with `mount-buildkite-agent: false` and the old `/usr/bin/buildkite-agent:/usr/bin/buildkite-agent` bind mount in `docker-compose.yml`, this migration can remove the bind mount while leaving the plugin opted out. The Docker Compose plugin defaults this option to `false`, so the agent would remain unavailable inside the container.

Suggested direction:
Mirror the docker plugin handling here: when the existing value is `false`, flip it to `true`, and add a test for the Docker Compose plugin false case.

Reference: https://github.com/buildkite-plugins/docker-compose-buildkite-plugin#mount-buildkite-agent-run-only-boolean

KSHITIZ6341 (commented):
One migration edge case that looks worth fixing before merge.

72636c (approved):
👏

PR #2490, opened 2026-06-23
Guard against autofix loops on Renovate lock file updates

72636c:
Perhaps not worth it, but I've already been down the rabbit hole...

https://seekchat.slack.com/archives/C06QWMV5BSA/p1782180054180849


src/cli/lint/autofix.ts
```diff
@@ -17,6 +19,11 @@ import * as Buildkite from '@skuba-lib/api/buildkite';
 import * as Git from '@skuba-lib/api/git';
 import * as GitHub from '@skuba-lib/api/github';
 
+export const RENOVATE_AUTHOR = {
+  name: 'renovate[bot]',
+  email: '29139614+renovate[bot]@users.noreply.github.com',
```
etaoins, line 24: If anyone else is curious, `29139614` is the GitHub user ID for Renovate

src/cli/lint/autofix.ts
```diff
@@ -97,6 +117,104 @@ const shouldPush = async ({
...
+const isRenovateLockfileUpdateInGit = async (
+  dir: string,
+): Promise<boolean | null> => {
+  const [headResult] = await git.log({ depth: 1, dir, fs });
+  if (!headResult) {
+    return null;
+  }
+
+  // Check whether the head commit was authored by Renovate.
+  // This likely isn't perfect when there are timing issues and/or another bot
```
etaoins, line 160: Probably the key thing here is we will eventually break the loop, not that we 100% break it on the first try

etaoins (approved):
I feel super bad that this is a one character change in the affected repo, but presumably this is something other repos can fall in to

samchungy (approved):

PR #2495, opened 2026-06-29
Update `eslint-config-seek dep to 15.0.6`

eddeee888:
Skuba uses `vitest/valid-title` rule [here](https://github.com/seek-oss/skuba/blob/65b90382f2e81347b3516713fac1c2c9a85b78b1/packages/eslint-config-skuba/src/index.ts#L74) which is available in later versions of `@vitest/eslint-plugin` (at least >1.6.6)

However, `eslint-config-seek@15.0.5` was only bringing in at the minimum `^1.5.2`.
This means some codebases could be stuck in earlier versions of `@vitest/eslint-plugin` without manual dep bump or lockfile refresh.

This PR bumps `eslint-config-seek` to `15.0.6` which brings in `@vitest/eslint-plugin^1.6.20` which should fix this issue.

### DIFF
--- a/packages/eslint-config-skuba/package.json
+++ b/packages/eslint-config-skuba/package.json
@@ -40,7 +40,7 @@
     "test:ci": "echo \"No tests\""
   },
   "dependencies": {
-    "eslint-config-seek": "15.0.5",
+    "eslint-config-seek": "15.0.6",
     "eslint-plugin-skuba": "workspace:*",
     "eslint-plugin-yml": "^3.0.0",
     "typescript-eslint": "^8.59.2"
--- a/.changeset/great-cougars-go.md
+++ b/.changeset/great-cougars-go.md
@@ -1,0 +1,5 @@
+---
+'skuba': patch
+---
+
+Update `eslint-config-seek` dep to `15.0.6`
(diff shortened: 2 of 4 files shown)


packages/eslint-config-skuba/package.json
```diff
@@ -40,7 +40,7 @@
     "test:ci": "echo \"No tests\""
   },
   "dependencies": {
-    "eslint-config-seek": "15.0.5",
+    "eslint-config-seek": "15.0.6",
```
72636c, line 43: Are we intentionally pinning this @samchungy?

samchungy: I don't remember 🤔

72636c: I will caret, this looks stale: https://github.com/seek-oss/skuba/blob/c05fb14c55779f4a57341a2f385c2e747ee767f8/.changeset/plenty-cameras-drum.md

eddeee888: Would you like me to do it as part of this PR? Or you've got it? 👀

samchungy: Ryan did it in a different PR 🚀

PR #2501, opened 2026-06-29
Restore checkrun

samchungy:
Upon request from a few teams. I'm waiting for Vitest 5 to come out before creating annotations once again


zbrydon (approved):

src/cli/test/index.ts
```diff
@@ -7,5 +13,17 @@ export const test = async () => {
     preferLocal: false,
   });
 
-  return customExec('vitest', ...argv);
+  let result: { code: number } | undefined;
+  try {
+    result = await customExec('vitest', ...argv);
+  } catch {
+    process.exitCode = 1;
+  }
```
72636c, line 21: Is it worth re-throwing the caught error at the end, or does Vitest always clearly write the failure itself?

72636c (approved):

src/cli/test/index.ts
samchungy: image

I think the errors are pretty clear without

PR #2513, opened 2026-07-09
Bundle skuba-doo kinda

samchungy:
Probably makes more sense to rely `build-package` rather than `build` anyway.

But I wanna 🗑️ the esbuild at some point so this is a good excuse to move it to use `build-package`


scripts/test-template.sh
```diff
@@ -55,8 +55,9 @@ awk '/^[^ \t#]/ { skip = ($0 ~ /^packages:/ || $0 ~ /^overrides:/) } !skip { pri
 
 cd "../${skuba_temp_directory}" || exit 1
 
-echo "pnpm init"
-pnpm init
```
samchungy, line 59: On my local because we rely on package.json for the pnpm version tries to use pnpm 11 lol

tsdown.config.mts
```diff
@@ -0,0 +1,36 @@
...
+    exports: {
+      bin: 'src/skuba.ts',
+      exclude: ['!index'],
+      customExports: {
+        './config/eslint.js': './config/eslint.js',
+        './config/prettier.js': './config/prettier.js',
+        './config/tsconfig.json': './config/tsconfig.json',
+      },
+    },
+    entry: [
```
samchungy, line 25: Can maybe look to refactor some of this setup down the line. We currently rely on things like the following:

https://github.com/seek-oss/skuba/blob/1b99ddaabe06edd02d70b090f97d9cd1ec02fe21/src/utils/command.ts#L12-L13

Which aren't too bundler friendly.

tsdown.config.mts
```diff
@@ -0,0 +1,36 @@
+import { defineConfig } from 'tsdown/config';
+
+export default defineConfig([
+  {
+    entry: ['src/index.ts'],
+    exports: true,
+    format: ['esm'],
+    outDir: 'lib',
+    dts: true,
```
samchungy, line 13: This is separated so we can generate a .dts specifically for this file and avoid generating one for the other files

72636c (dismissed):
no

package.json
```diff
@@ -126,6 +130,7 @@
     "read-package-up": "^12.0.0",
     "rolldown": "^1.1.0",
     "semantic-release": "^25.0.2",
+    "semver": "7.8.0",
```
samchungy: This seems to be missed as a dep

tsdown.config.mts
zbrydon, line 1: do we want to throw in `attw` and `publint` for good measure?

package.json
zbrydon: Should this be pinned?

zbrydon (approved):

package.json
samchungy: no

PR #2534, opened 2026-07-24
Fix `.npmrc` secret guardrail

zbrydon:


src/cli/lint/autofix.ts
```diff
@@ -123,9 +115,29 @@ const createAutofixIgnore = async ({
...
+  const gitRoot = await Git.findRoot({ dir });
+
+  const allChangedFiles = await Git.getChangedFiles({
+    dir,
+    ignore: AUTOFIX_IGNORE_FILES_BASE,
+  });
+
+  const npmrcSecretIgnores = gitRoot
+    ? await getNpmrcSecretIgnores(gitRoot, allChangedFiles)
+    : [];
```
72636c: Is there any value in falling back to the `dir`?

(FWIW I think we require Git root to resolve in order to push autofixes on some other code path 🍝)

src/cli/lint/autofix.ts
```diff
@@ -123,9 +115,29 @@ const createAutofixIgnore = async ({
   currentBranch?: string;
   dir: string;
 }): Promise<Git.ChangedFile[] | false> => {
-  const ignore = await getIgnores(dir);
+  const gitRoot = await Git.findRoot({ dir });
+
+  const allChangedFiles = await Git.getChangedFiles({
```
72636c: Perhaps `unsafeChangedFiles`?

src/cli/lint/autofix.ts
```diff
@@ -215,15 +227,21 @@ const isRenovateLockfileUpdate = async ({
...
+  const npmrcChanges = changedFiles.filter((file) => isNpmrc(file.path));
+
+  const results = await Promise.all(
+    npmrcChanges.map(async (file) =>
+      hasNpmrcSecret((await readFile(file.path)) ?? '') ? file : null,
+    ),
+  );
 
-  return AUTOFIX_IGNORE_FILES_BASE;
+  return results.filter((file): file is Git.ChangedFile => file !== null);
```
72636c: Do we need the type guard?

```suggestion
  return results.filter((file) => file !== null);
```

72636c (approved):
Ace, thank you

src/cli/lint/autofix.ts
zbrydon: No 🙃

src/cli/lint/autofix.ts
zbrydon: yeah true but falling back to it is probs better than doing nothing in case for whatever reason the requirement for a Git root changes at some point

PR #2564, opened 2026-09-01
Migrate to @clack/prompts

samchungy:
Zac said it was pretty

image


src/cli/init/index.ts
```diff
@@ -34,6 +40,133 @@ import { writePackageJson } from './writePackageJson.js';
...
+  onLine: (line: string) => void,
+) => {
+  if (!readable) {
+    return;
+  }
+
+  readline.createInterface({ input: readable }).on('line', onLine);
+};
+
+const createTaskLogLogger = (
```
samchungy, line 54: A little gross but essentially shimming our logger interface to pass into our prettier implementation

src/cli/init/prompts.ts
```diff
@@ -1,145 +1,241 @@
...
   name: string;
   message: string;
   initial?: string;
-  validate?: (value: string) => boolean | string | Promise<boolean | string>;
-  /**
-   * Whether the user is allowed to skip field entry and use the initial value.
-   *
-   * Defaults to `false`.
-   */
-  allowInitial?: boolean;
```
samchungy, line 26: AllowInitial was dropped in favour of clack's native default/placeholder functionality

mrjono1 (approved):
it looks nice

src/cli/init/prompts.ts
72636c: Ace. Not that we have many templates nor document the contract well, but are we comfortable that `initial` maps to `placeholder` and there is no option for `defaultValue`?

https://github.com/seek-oss/skuba/blob/a13216fdae6947ae196d2f9757aa29fe5e36f292/src/utils/template.ts#L63-L76

72636c (approved):

src/cli/init/prompts.ts
samchungy: I did do a search of all the repos and couldn't see `allowInitial` actually used. I figure we can just add a default option if it comes up

I think given the UI Clack gives us it makes sense to let the initial be the default + placeholder? 
image

72636c: From https://github.com/search?q=org%3ASEEK-Jobs+path%3A**%2Fskuba.template.js+%22initial%22&type=code, existing `initial`s are definitely `placeholder`s and not `defaultValue`s, so happy to keep this as just the former for now 👍

